PR #15, opened 2021-01-12
Additions to files and jobs. 

ToineMichielseBroadcom:
Hello Guilherme

Not sure if you are working on a more permanent solution to manage the request headers. I just left it... (well added the ZOSMD header in the SdkApi as well as the submit text as jcl method.

In Files:
- added method to get list of USS file similar to zOS datasets

In Jobs:
- Added various methods, most important a simple to use method to get all the output as files in a specified directory. This will help greatly in DevOps type scripts.

I have also create a few sample python scripts.


ToineMichielseBroadcom:
Fixing signoff

PR #17, opened 2021-01-13
This will add methods to files.py and jobs.py as well as a directory with python samples

ToineMichielseBroadcom:


ToineMichielseBroadcom:
Generated gpg key. will regenerate commits

PR #21, opened 2021-01-19
zos_files: allow to specify pattern, limit and start in list_dsn_members

jguillaumes:
I've added functionality to the list_dsn_members of zos_files to:
- Filter the members by pattern.
- Specify the limit of returned names. ZOSMF defaults to 1000, and so does this method.
- Specifiy the first member to return, in case you need to do multiple calls to retrieve all members.

Rationale: we've got libraries with literally hundreds of thousands of members, and to be able to use this API we need the added functionality.

Unless I've made a big mistake, the code already written using this class sholuld not be affected. The new parameters have default values corresponding to the current functionality.


gcartier94:
Jordi thank you for your contribution,

Apologizes for the late response, as you may know, the Zowe Python SDK is undergoing some architectural changes (described [here]). These changes prevent me from accepting your contributions to the main/master branch as they might conflict. However, your contributions are valuable and an important step into establishing the Python SDK as a more mature framework. Therefore your contributions will be accepted without any changes to the new `experimental` branch. 

The `experimental` branch will consolidate new features that have not yet been fully tested and/or integrated into the project but can be leveraged by the community.

When the architectural changes are finished, your contributions will be merged into the main code. Some change might be required , but I'll do the necessary adjustments when the time comes

PR #38, opened 2022-03-18
feat: added write_to_uss and upload_file_to_uss methods

### LINKED ISSUE #37: [zos-files] Add method to upload file to USS
For this enhancement, we probably want two new methods:
- `upload_file_to_uss` modeled after `upload_file_to_dsn`:
  https://github.com/zowe/zowe-client-python-sdk/blob/4a413309c144a629bb0e87b7a6fb659a91923e30/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py#L194-L201
- `write_to_uss` to call the `requests` library and write data to the USS file using the REST endpoint documented [here](https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-write-data-zos-unix-file#PutWriteUnixFile)

himanshu007-creator:
resolves #37

### DIFF
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
@@ -208,7 +208,7 @@ class Files(SdkApi):
             A JSON containing the result of the operation
         """
         custom_args = self.create_custom_request_arguments()
-        custom_args["url"] = "{}fs/{}".format(self.request_endpoint, filepath_name)
+        custom_args["url"] = "{}fs/{}".format(self.request_endpoint, filepath_name.lstrip("/"))
         custom_args["data"] = data
         custom_args['headers']['Content-Type'] = 'text/plain; charset={}'.format(encoding)
         response_json = self.request_handler.perform_request(
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
@@ -199,3 +199,28 @@ class Files(SdkApi):
             response_json = self.write_to_dsn(dataset_name, file_contents)
         else:
             raise FileNotFound(input_file)
+
+    def write_to_uss(self, filepath_name, data, encoding=_ZOWE_FILES_DEFAULT_ENCODING):
+        """Write content to an existing UNIX file.
+        Returns
+        -------
+        json
+            A JSON containing the result of the operation
+        """
+        custom_args = self.create_custom_request_arguments()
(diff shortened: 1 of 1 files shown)


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -199,3 +199,28 @@ def upload_file_to_dsn(self, input_file, dataset_name, encoding=_ZOWE_FILES_DEFA
...
+
+    def write_to_uss(self, filepath_name, data, encoding=_ZOWE_FILES_DEFAULT_ENCODING):
+        """Write content to an existing UNIX file.
+        Returns
+        -------
+        json
+            A JSON containing the result of the operation
+        """
+        custom_args = self.create_custom_request_arguments()
+        custom_args["url"] = "{}fs/{}".format(self.request_endpoint, filepath_name)
```
t1m0thyj, line 211: If `filepath_name` contains a leading slash, then the string generated here will contain two consecutive slashes and the z/OSMF API doesn't like that.

I believe we could fix this by removing the leading slash if present - something like `filepath_name.lstrip("/")`

himanshu007-creator: thanks !

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -199,3 +199,28 @@ def upload_file_to_dsn(self, input_file, dataset_name, encoding=_ZOWE_FILES_DEFA
...
             raise FileNotFound(input_file)
+
+    def write_to_uss(self, filepath_name, data, encoding=_ZOWE_FILES_DEFAULT_ENCODING):
+        """Write content to an existing UNIX file.
+        Returns
+        -------
+        json
+            A JSON containing the result of the operation
+        """
+        custom_args = self.create_custom_request_arguments()
```
t1m0thyj, line 210: `self.create_custom_request_arguments` -> `self.__create_custom_request_arguments`

t1m0thyj (commented):
@himanshu007-creator Thanks for your contribution!

I've updated the target branch of your pull request to the newly created `next` branch. This branch contains enhancements from both the `develop` and `experimental` branches, including unit and integration tests.

Could you please address the comment below, and add a test in [test_zos_files.py](https://github.com/zowe/zowe-client-python-sdk/blob/next/tests/integration/test_zos_files.py)? The test can be modeled after `test_write_to_dsn_should_be_possible`, but for USS (`write_to_uss`) instead.

himanshu007-creator:
Thanks ! @t1m0thyj  🙂

PR #44, opened 2022-03-22
Stream requests

richardgreg:


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -186,16 +225,15 @@ def download_binary_dsn(self, dataset_name, output_file, with_prefixes=False):
...
         out_file.close()
 
     def upload_file_to_dsn(self, input_file, dataset_name, encoding=_ZOWE_FILES_DEFAULT_ENCODING):
         """Upload contents of a given file and uploads it to a dataset."""
         if os.path.isfile(input_file):
-            in_file = open(input_file, 'r')
-            file_contents = in_file.read()
-            response_json = self.write_to_dsn(dataset_name, file_contents)
+            with open(input_file, 'r') as file_contents:
+                response_json = self.write_to_dsn(dataset_name, file_contents)
```
t1m0thyj, line 237: Using the variable name `in_file` would be better here instead of `file_contents`, since we are no longer reading the contents of the file but now passing the file object on to `requests`.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -146,6 +160,31 @@ def get_dsn_binary_content(self, dataset_name, with_prefixes=False):
...
+        custom_args = self.create_custom_request_arguments()
+        custom_args["url"] = "{}ds/{}".format(self.request_endpoint, dataset_name)
+        custom_args["headers"]["Accept"] = "application/octet-stream"
+        if with_prefixes:
+            custom_args["headers"]["X-IBM-Data-Type"] = 'record'
+        else:
+            custom_args["headers"]["X-IBM-Data-Type"] = 'binary'
+        custom_args["stream"] = True
+        content = self.request_handler.perform_request("GET", custom_args)
+        return content
```
t1m0thyj, line 186: In addition to setting `stream = True`, the `perform_request` method needs to be changed to use a method that returns a stream instead of a string. See comment above.

t1m0thyj (changes requested):

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -122,6 +122,20 @@ def get_dsn_content(self, dataset_name):
...
+        """Retrieve the contents of a given dataset streamed.
+
+        Returns
+        -------
+        json
+            A JSON with the contents of a given dataset
+        """
+        custom_args = self.create_custom_request_arguments()
+        custom_args["url"] = "{}ds/{}".format(self.request_endpoint, dataset_name)
+        custom_args["stream"] = True
```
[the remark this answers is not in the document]
richardgreg: What about adding this line to `request_handler.normalize_response`
```python
if 'stream' in self.request_arguments and self.request_arguments['stream'] == True:
    return self.response.raw
```
Then in `download_dsn`
```python
raw_response = self.get_dsn_content_streamed(dataset_name)
out_file = open(output_file, 'w')
shutil.copyfileobj(raw_response, out_file)
out_file.close()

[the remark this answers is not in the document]
Comment (t1m0thyj) [reply src/zos_files/zowe/zos_files_for_zowe_sdk/files.py:135]: Good idea for `request_handler.normalize_response`, I think that will be nicer than creating a separate `_streamed` method 👍 

In `download_dsn` instead of calling `out_file.close`, I'd suggest using a `with` block:
```
with open(output_file, 'w') as out_file:
    shutil.copyfileobj(raw_response, out_file)
```

**Comment by richardgreg** [comment]
Style-wise, is it okay if I squash all commit messages into one, I'll note the changes in the body of the commit message? Or additional modifications should have a new commit?

**Comment by t1m0thyj** [comment]
> Style-wise, is it okay if I squash all commit messages into one, I'll note the changes in the body of the commit message? Or additional modifications should have a new commit?

@richardgreg I'd recommend making additional modifications in a new commit, as it's typically easier to work that way both for the person making the changes and the PR reviewer. 🙂 In this case, the diff is small enough that squashing all commits would also be fine if you prefer.

File: src/zos_files/zowe/zos_files_for_zowe_sdk/files.py [hunk 4]
```diff
@@ -122,6 +123,20 @@ def get_dsn_content(self, dataset_name):
...
 
+    def get_dsn_content_streamed(self, dataset_name):
+        """Retrieve the contents of a given dataset streamed.
+
+        Returns
+        -------
+        json
+            A JSON with the contents of a given dataset
+        """
+        custom_args = self.create_custom_request_arguments()
```
t1m0thyj, line 134: `self.create_custom_request_arguments` -> `self.__create_custom_request_arguments`

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -146,6 +161,31 @@ def get_dsn_binary_content(self, dataset_name, with_prefixes=False):
...
+        ----------
+        dataset_name: str - Name of the dataset to retrieve
+        with_prefixes: boolean - if True include a 4 byte big endian record len prefix
+                                 default: False 
+        Returns
+        -------
+        bytes
+            The contents of the dataset with no transformation
+        """
+        custom_args = self.create_custom_request_arguments()
```
t1m0thyj, line 178: `self.create_custom_request_arguments` -> `self.__create_custom_request_arguments`

t1m0thyj (commented):
@richardgreg Thanks for your contribution!

I've updated the target branch of your pull request to the newly created `next` branch. This branch contains enhancements from both the `develop` and `experimental` branches, including unit and integration tests.

Could you please address the comments below and add a test in [test_zos_files.py](https://github.com/zowe/zowe-client-python-sdk/blob/next/tests/integration/test_zos_files.py)? The test can be modeled after `test_get_dsn_content_should_return_content_from_dataset`, but for streamed data sets (`get_dsn_content_streamed`) instead.

richardgreg:
Is there a doc regarding running tests?
I'm having issues when running the test file. I get FileNotFoundError: [Errno 2] No such file or directory: '~/.zowe/profiles/zosmf/.yaml'

t1m0thyj:
> Is there a doc regarding running tests? I'm having issues when running the test file. I get FileNotFoundError: [Errno 2] No such file or directory: '~/.zowe/profiles/zosmf/.yaml'

The `setUp` stage of the tests in test_zos_files.py expects that a `ZOWE_TEST_PROFILE` environment variable is defined to specify the name of a Zowe CLI z/OSMF profile to be used for the tests.

This z/OSMF profile would live in `~/.zowe/profiles/zosmf/<profileName>.yaml` and can be created following the documentation here: https://docs.zowe.org/stable/user-guide/cli-usingcli#using-profiles

Without access to a z/OSMF instance running on the mainframe, a z/OSMF profile won't be very useful and the tests will still fail. If you don't have access to one, I can run the integration tests in your branch to verify they pass 🙂

PR #48, opened 2022-05-19
Added method to delete a file in Files Class

shalearkane:
Added delete_uss method to delete a file by sending the file path to this endpoint : /zosmf/restfiles/fs/<file-pathname> as mentioned on https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-delete-unix-file-directory#DeleteUnixFile .

On successful deletion, the method receives HTTP Response Code 204.

### DIFF
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
@@ -68,7 +68,7 @@ class Files(SdkApi):
         response_json = self.request_handler.perform_request("GET", custom_args)
         return response_json
 
-        def delete_uss(self, filepath_name):
+    def delete_uss(self, filepath_name):
         """
         Delete a file or directory
 
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
@@ -39,7 +39,7 @@ class Files(SdkApi):
         """
         super().__init__(connection, "/zosmf/restfiles/")
 
-        
+
     def list_files(self, path):
         """Retrieve a list of USS files based on a given pattern.
 
@@ -53,7 +53,7 @@ class Files(SdkApi):
         custom_args["url"] = "{}fs".format(self.request_endpoint)
         response_json = self.request_handler.perform_request("GET", custom_args)
         return response_json
-        
+
     def get_file_content(self, filepath_name):
         """Retrieve the content of a filename. The complete path must be specified.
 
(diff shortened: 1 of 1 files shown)


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -36,10 +36,13 @@ def __init__(self, connection):
         ----------
         connection
             The z/OSMF connection object (generated by the ZoweSDK object)
+
+        Also update header to accept gzip encoded responses
         """
         super().__init__(connection, "/zosmf/restfiles/")
+        self.default_headers["Accept-Encoding"] = "gzip"
```
t1m0thyj, line 43: I like the approach of adding the header in a common function so we don't have to do it in multiple places 👍 

Do you know if the Python `requests` library automatically decompresses REST responses that have the `Content-Encoding: gzip` header defined? If it doesn't, there will be more changes required to decode compressed responses.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -68,6 +71,24 @@ def get_file_content(self, filepath_name):
...
+        ----------
+        filepath of the file to be deleted
+
+        Returns
+        -------
+        204
+            HTTP Response for No Content
+        """
+        custom_args = self.create_custom_request_arguments()
+        custom_args["url"] = "{}fs{}".format(self.request_endpoint,filepath_name)
```
t1m0thyj, line 88: The `filepath_name` should be supported either with or without a leading slash. Could you change it to something like this:
```custom_args["url"] = "{}fs/{}".format(self.request_endpoint, filepath_name.lstrip("/"))```

File: src/zos_files/zowe/zos_files_for_zowe_sdk/files.py [hunk 3]
```diff
@@ -68,6 +71,24 @@ def get_file_content(self, filepath_name):
...
+        response_json = self.request_handler.perform_request("DELETE", custom_args)
```
t1m0thyj, line 89: Thanks for documenting the expected response code of 204 👍 It also needs to be specified on the `perform_request` method with the argument `expected_code=[204]` in order for the request to succeed with a non-200 response code.

t1m0thyj (changes requested):
Thanks for your contribution! Left a few comments.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
shalearkane: Yes, according to the FAQ, it does https://2.python-requests.org/en/master/community/faq/#encoded-data

However,  this issue on github https://github.com/psf/requests/issues/2446 , and
 https://2.python-requests.org/en/master/user/quickstart/#binary-response-content 
mentions that only those responses are decoded which have headers `Transfer-Encoding: gzip`

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
shalearkane: Yaa, thanks for the review. I have made the necessary changes to the code.

Issue #51, opened 2022-05-23
Create GitHub workflows


t1m0thyj:
It would be nice for this repo to have fully automated CICD with GitHub Actions 🙂 Ideas for workflows:
* [x] Run unit tests and report coverage
* [x] Perform linting and style checks
* [x] Deploy package updates to PyPI
* [ ] Run system tests in Marist environment

aadityasinha-dotcom:
Can I work on this?

zFernand0:
Closing in favor of:
- #261

PR #87, opened 2022-06-17
Set up automated coverage check and document cmd to run unit tests

### LINKED ISSUE #39: Add test suites
There are some tests in the [develop](https://github.com/zowe/zowe-client-python-sdk/tree/develop/tests) branch that we can model after.

* [x] Unit tests to test individual methods
* [ ] System tests against a z/OSMF instance
* [x] GitHub workflow to automate tests (related to #51)

richardgreg:
Fixes #39 #51

Signed-off-by: Richard Gregory <richardgrecoson@gmail.com>

### DIFF
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -3,11 +3,7 @@
 
 name: Zowe SDK Build
 
-on:
-  push:
-    branches: [ develop ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


t1m0thyj:
To test out the coverage check, could you update these lines in the workflow: https://github.com/zowe/zowe-client-python-sdk/blob/960888b527ec9b2dd4871742d3216e0ed7b19845/.github/workflows/sdk-build.yml#L6-L10

For now I'd suggest running on all push and pull request events, so the `on` clause could be replaced with:
```
on: [push, pull_request]
```

Then the workflow should run on this PR and we can see whether the coverage results get uploaded on Codecov.io 🙂

richardgreg:
Do we need a secret token?

.github/workflows/sdk-build.yml
```diff
@@ -32,4 +28,14 @@ jobs:
...
       run: |
-        green -vvv ./tests/unit
\ No newline at end of file
+        green -vvv -r ./tests/unit
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
+      with:
+        directory: ./
+        fail_ci_if_error: true
+        files: ./coverage
```
t1m0thyj, line 37: @richardgreg A secret token shouldn't be necessary for public repositories. I think the issue is related to the `files` defined here (see the Codecov error below):
```
[2022-06-18T15:[39](https://github.com/zowe/zowe-client-python-sdk/runs/6949083223?check_suite_focus=true#step:7:40):32.958Z] ['error'] None of the following appear to exist as files: ./coverage
[2022-06-18T15:39:32.959Z] ['error'] There was an error running the uploader: Error while cleaning paths. No paths matched existing files!
```

Should the path be `./.coverage` (with an extra `.`) instead of `./coverage`?

Issue #81, opened 2022-06-15
[zos-files] `upload_file_to_dsn` fails using streaming


t1m0thyj:
The following sample code throws an error.
```python
my_files = Files(connection)
print(my_files.upload_file_to_dsn("samples/jcl/sleep.jcl", connection["user"] + ".PUBLIC.JCL(SLEEP)"))
```
![image](https://user-images.githubusercontent.com/22344007/173858657-5de13152-7745-42b7-869a-98fa9fe79c9d.png)

Issue #82, opened 2022-06-15
[zos-files] `download_dsn` fails using streaming


t1m0thyj:
The following sample code throws an error:
```python
my_files = Files(connection)
print(my_files.download_dsn(connection["user"] + ".PUBLIC.JCL(SLEEP)", "test.txt"))
```
![image](https://user-images.githubusercontent.com/22344007/173887359-0a4f7831-ee8e-4e83-8233-3abdac3035ab.png)

`download_binary_dsn` also fails with the same error.

PR #92, opened 2022-06-28
Fix streaming requests failing

richardgreg:
Also added documentation running unit tests

Fixes: #81 #82


richardgreg:
Spent a fair amount time trying to refactor but that just created a whole bunch of mess 😆. I'm open to ideas

CONTRIBUTING.md
```diff
@@ -32,6 +32,16 @@ The project's test suite can be run with the python test runner, `green`
 green -vvv  ./tests/unit
 ```
 
+In order to run the integration test, one will need to have a zowe profile. If zowe cli is installed, the test profile lives in `~/.zowe/profiles/zomsf/Zowe.yaml`
+
+```
+host: zzow03.zowe.marist.cloud
+port: 10443
```
t1m0thyj, line 39: This host and port is specific to the LFX mentorship.

For the documentation we should use a more generic host and port, something like `example.com` and `443`.

src/core/zowe/core_for_zowe_sdk/request_handler.py
```diff
@@ -72,6 +72,30 @@ def perform_request(self, method, request_arguments, expected_code=[200]):
...
+
+        Returns
+        -------
+        A raw response data
+        """
+        self.method = method
+        self.request_arguments = request_arguments
+        self.expected_code = expected_code
+        self.__validate_method()
+        self.__send_stream_request()
```
t1m0thyj, line 95: Would it make sense to add an optional `stream=True` parameter on the existing `__send_request` method?

CONTRIBUTING.md
```diff
@@ -32,6 +32,16 @@ The project's test suite can be run with the python test runner, `green`
 green -vvv  ./tests/unit
 ```
 
+In order to run the integration test, one will need to have a zowe profile. If zowe cli is installed, the test profile lives in `~/.zowe/profiles/zomsf/Zowe.yaml`
```
t1m0thyj, line 35: The profile doesn't have to be called "Zowe.yaml", the name can be anything as long as it matches the `ZOWE_TEST_PROFILE` variable defined in your `.env` file.

Could you mention configuring the `.env` file here as well? To run the integration tests, it is necessary to define `ZOWE_TEST_PROFILE='<myProfile>'` in the `.env` file, so that tests will load the profile from `~/.zowe/profiles/zosmf/<myProfile>.yaml`.

docs/source/testing/testing.rst
```diff
@@ -0,0 +1,19 @@
+Running Tests
```
t1m0thyj, line 1: Thanks for adding the doc here as well! Same comments apply as to the `CONTRIBUTING.md` file.

src/core/zowe/core_for_zowe_sdk/request_handler.py
```diff
@@ -72,6 +72,30 @@ def perform_request(self, method, request_arguments, expected_code=[200]):
...
+        -------
+        A raw response data
+        """
+        self.method = method
+        self.request_arguments = request_arguments
+        self.expected_code = expected_code
+        self.__validate_method()
+        self.__send_request(stream=True)
+        self.__validate_response()
+        return self.response.raw
```
t1m0thyj, line 97: In order to reach 80% coverage on the Codecov patch check, could you add a unit test for the `perform_streamed_request` method?

There is already a unit test suite for `RequestHandler` where it could be added: https://github.com/zowe/zowe-client-python-sdk/blob/next/tests/unit/test_zowe_core.py#L59-L69

A suggestion for a unit test would be to mock the `__send_request` method (see [here](https://stackoverflow.com/a/47407845/5504760) for an example of mocking a method inside a class), and assert that it is called by `perform_streamed_request`.

richardgreg: An issue I face with mocking `send_request` is that it seems because `perform_streamed_request` calls `send_request` it causes an error

t1m0thyj: I think the issue with mocking `__send_request` is that it cannot be mocked because of the double underscores in front which makes it a private method. It may work to mock `requests.Session.send` instead, since it is a public method and mocking it should stop an actual REST request from being made.

tests/unit/test_zowe_core.py
```diff
@@ -68,6 +69,14 @@ def test_object_should_be_instance_of_class(self):
         request_handler = RequestHandler(self.session_arguments)
         self.assertIsInstance(request_handler, RequestHandler)
 
+    @mock.patch('requests.Session.send')
+    def test_perform_streamed_request(self, mock_send_request):
+        """Performing a streamed request should call 'send_request' method"""
+        mock_send_request.return_value = mock.Mock(status_code=200)
+        request_handler = RequestHandler(self.session_arguments)
+        request_handler.perform_streamed_request("GET", {'url': 'https://www.zowe.org'})
+        mock_send_request.assert_called_once()
```
t1m0thyj, line 78: Thanks for adding the unit test, now the patch coverage for this PR is >80% 🎉 

In the unit tests, we want to validate that the important behavior of methods works as expected. In this case, the `perform_streamed_request` method needs to pass `stream=True` when it sends a request.

To test this, I would suggest adding another assertion at the end of the test:
`self.assertTrue(mock_send_request.call_args.kwargs["stream"])`
And if the assertion fails, then please fix the behavior of the `perform_streamed_request` method so it passes 🙂

t1m0thyj: Oops, I didn't realize that `call_args.kwargs` only exists in Python 3.8 and newer.

Replacing it with `call_args[1]` should work in older versions of Python, like this:
`self.assertTrue(mock_send_request.call_args[1]["stream"])`

t1m0thyj (approved):

PR #94, opened 2022-07-04
ProfileManager

shalearkane:
Added ProfileManager Class to manage Team Configs.


src/core/zowe/core_for_zowe_sdk/zosmf_profile2.py
```diff
@@ -0,0 +1,300 @@
+"""Zowe Python Client SDK.
```
t1m0thyj, line 1: Please rename the file to match the class it defines - I think `profile_manager.py` would make sense.

PR #96, opened 2022-07-10
Added Profile Manager Class

shalearkane:
Signed-off-by: Soumik Dutta <shalearkane@gmail.com>


src/core/zowe/core_for_zowe_sdk/zosmf_profile2.py
```diff
@@ -0,0 +1,302 @@
...
+            path = os.path.join(self._config_dir, filename)
+
+            if os.path.isfile(path):
+                self._config_filename = filename
+            else:
+                raise FileNotFoundError(
+                    f"file not found on path {path} or set directory path first"
+                )
+        else:
+            self._config_filename = filename
```
t1m0thyj, line 88: Instead of the `config_filename` getter/setter, I would suggest the following:
* The constructor can take an `appname` argument that defaults to `"zowe"` (`def __init__(self, appname="zowe"):`)
* `config_appname` can be a read-only property that returns the app name given to the constructor
* `config_filename` can be a read-only property like `config_filepath` that returns `{self.appname}.config.json`
* (in a future PR) `config_user_filename` can be a read-only property that returns `{self.appname}.config.user.json`

These behaviors match the Config class in the Node.js SDK:
https://github.com/zowe/imperative/blob/master/packages/config/src/Config.ts#L52

t1m0thyj (changes requested):
Overall this PR looks great! Thanks for adding thorough tests and documentation.

Left a few comments. Will review this PR again once failing tests have been resolved.

src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -11,48 +11,44 @@
...
-        The default endpoint for the API
     """
 
-    def __init__(self, connection, default_url):
-        if "plugin_profile" in connection:
-            self.connection = ZosmfProfile(connection['plugin_profile']).load()
-        else:
-            self.connection = ApiConnection(**connection)
+    def __init__(self, default_url, session):
+        self.session: ISession = session
```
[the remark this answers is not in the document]
richardgreg: For sure

src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -11,48 +11,44 @@
...
         self.default_service_url = default_url
         self.default_headers = {
             "Content-type": "application/json",
-            "X-CSRF-ZOSMF-HEADER": ""
+            "X-CSRF-ZOSMF-HEADER": "",
         }
+
         self.request_endpoint = "https://{base_url}{service}".format(
-            base_url=self.connection.host_url, service=self.default_service_url
+            base_url=Session.host_url, service=self.default_service_url
```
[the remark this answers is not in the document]
richardgreg: There's a `Session.host_url` property does contains that. If the session class is imported into the sdkapi, it should work right?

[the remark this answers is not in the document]
t1m0thyj: Yes, the `Session.host_url` property contains the information we want 🙂 Since it has an `@property` decorator, we need to access the `host_url` of a session instance, not the `host_url` of the static Session class.

If you store an instance of the Session class earlier:
```
        session = Session(profile)
        self.session: ISession = session.load()
```

Then you should be able to generate the request endpoint like this:
```
        self.request_endpoint = session.host_url + self.default_service_url
```

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass
+class ISession:
+    """
+    Class to represent session parameters
+    """
+
+    hostname: str
+    port: int
+    rejectUnauthorised: bool = True
```
t1m0thyj, line 28: Please change the spelling to "rejectUnauthorized" in all places where this is used.

I know this is a difference between British and American spelling, but we should be consistent with existing Zowe code that expects it to be spelled a particular way 🙂

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,71 @@
...
+                hostname=props["host"], port=props["port"]
+            )
+        else:
+            raise "Hostname and Port must be supplied"
+
+        # determine authentication type
+        if props["user"] is not None and props["password"] is not None:
+            self.session.user = props["user"]
+            self.session.password = props["password"]
+            self.session.rejectUnauthorised = props["rejectUnauthorised"]
```
t1m0thyj, line 57: The "rejectUnauthorized" property may be set for any authentication type and may not always be defined in the `props` dictionary.

To load the "rejectUnauthorized" value, I would suggest adding an `if` condition outside of the user/password one. Something like the following:
```
if props.get("rejectUnauthorized") is not None:
    self.session.rejectUnauthorized = props["rejectUnauthorized"]
```

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class Session:
+    """
+    Class used to represent connection details
+    """
+
+    def __init__(self, props: dict) -> None:
+        # set hostname and port
+        if props["host"] is not None and props["port"] is not None:
```
t1m0thyj, line 46: `if props["host"] is not None and props["port"] is not None:`
will throw an error if either property is missing.
Instead of `props["x"] is not None`, please use `props.get("x") is not None`.

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,71 @@
...
+            self.session: ISession = ISession(
+                hostname=props["host"], port=props["port"]
+            )
+        else:
+            raise "Hostname and Port must be supplied"
+
+        # determine authentication type
+        if props["user"] is not None and props["password"] is not None:
```
t1m0thyj, line 54: (same comment as above)

`if props["user"] is not None and props["password"] is not None:`
will throw an error if either property is missing.
Instead of `props["x"] is not None`, please use `props.get("x") is not None`.

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,71 @@
...
+            self.session.user = props["user"]
+            self.session.password = props["password"]
+            self.session.rejectUnauthorised = props["rejectUnauthorised"]
+            self.session.type = session_constants.AUTH_TYPE_BASIC
+        elif props["tokenType"] is not None and props["tokenValue"] is not None:
```
t1m0thyj, line 59: (same comment as above)

`if props["tokenType"] is not None and props["tokenValue"] is not None:`
will throw an error if either property is missing.
Instead of `props["x"] is not None`, please use `props.get("x") is not None`.

src/core/zowe/core_for_zowe_sdk/zosmf_profile2.py
```diff
@@ -0,0 +1,302 @@
...
+        try:
+            secure_config = keyring.get_password(
+                constants["ZoweServiceName"], constants["ZoweUsername"]
+            )
+        except Exception as exc:
+            raise SecureProfileLoadFailed(
+                constants["ZoweServiceName"], error_msg=str(exc)
+            ) from exc
+
+        secure_config_json = jsonc.loads(base64.b64decode(secure_config).decode())
```
t1m0thyj, line 282: This code fails on Windows where there is some platform-specific logic required for loading credentials:
https://github.com/zowe/zowe-client-python-sdk/blob/c2725e47555f99ba46edbc5d380ad38a6df99ca0/src/core/zowe/core_for_zowe_sdk/zosmf_profile.py#L100-L106

* The service name must have the account name added to the end like this: `{serviceName}/{accountName}`.
* The secret value must be decoded with UTF-16 before it can be Base64 decoded.

src/core/zowe/core_for_zowe_sdk/sdk_api.py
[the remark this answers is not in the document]
richardgreg: Understood

src/core/zowe/core_for_zowe_sdk/zosmf_profile2.py
shalearkane: hey, I've fixed the issue, could you test this on Windows?

t1m0thyj: Thanks, I can confirm this works now to load credentials on Windows.

I got a unit tests failure in `tests/unit/test_zowe_core.py` related to this. I believe adding the following snippet in that file after `SECURE_CONFIG_PROPS` is defined should fix it:
```
        if sys.platform == "win32":
            SECURE_CONFIG_PROPS = SECURE_CONFIG_PROPS.decode("utf-16")
```

We may also need to do further testing with the keyring on Linux but that could be handled in a separate issue. I forget what OS you're using, if you're on Linux and it's already working for you then maybe we're already covered 🙂

src/core/zowe/core_for_zowe_sdk/sdk_api.py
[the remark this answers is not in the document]
t1m0thyj: The `request_endpoint` is still not generated correctly because the protocol is being duplicated.

I would suggest changing it to use the code in the comment above. The advantage of doing it this way is:
```
session = Session(profile)
self.session: ISession = session.load()
```
lets us store both the Session class and the ISession dataclass for future use.

And then we can make use of the `Session.host_url` property without duplicating code for building the host URL string:
```
self.request_endpoint = session.host_url + self.default_service_url
```

src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -10,49 +10,45 @@
...
         self.session_arguments = {
-            "verify": self.connection.ssl_verification,
-            "timeout": 30
+            "verify": self.session.rejectUnauthorized,
+            "timeout": 30,
         }
         self.request_handler = RequestHandler(self.session_arguments)
 
+        if self.session.type == session_constants.AUTH_TYPE_BASIC:
+            self.request_arguments["auth"] = (self.session.user, self.session.user)
```
t1m0thyj, line 47: @richardgreg The second item in the tuple should be password not user.

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Class used to represent connection details
+    """
+
+    def __init__(self, props: dict) -> None:
+        # set host and port
+        if props.get("host") is not None:
+            self.session: ISession = ISession(host=props.get("host"))
+        else:
+            raise "host and Port must be supplied"
```
t1m0thyj, line 47: @shalearkane Since port is no longer required it shouldn't be mentioned here. This could say something like "Hostname must be supplied".

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,70 @@
...
+            self.session.type = session_constants.AUTH_TYPE_BASIC
+        elif props.get("tokenType") is not None and props.get("tokenValue") is not None:
+            self.session.tokenType = props.get("tokenType")
+            self.session.tokenValue = props.get("tokenValue")
+            self.session.type = session_constants.AUTH_TYPE_TOKEN
+        else:
+            raise "An authentication method must be supplied"
+
+        # set additional parameters
+        self.session.rejectUnauthorized = props.get("rejectUnauthorized", True)
```
t1m0thyj, line 63: @shalearkane There are more additional parameters to be set here: basePath, port, and protocol. Something like the following should work:
```
        self.session.basePath = props.get("basePath")
        self.session.port = props.get("port", self.session.port)
        self.session.protocol = props.get("protocol", self.session.protocol)
        self.session.rejectUnauthorized = props.get("rejectUnauthorized", self.session.rejectUnauthorized)
```

Base path does not have a default value like the other properties do.
Also I would suggest using the existing value on the session as a fallback to reduce code duplication (e.g., falling back to `self.session.rejectUnauthorized` instead of a hardcoded `true`).

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def load(self) -> ISession:
+        return self.session
+
+    @property
+    def host_url(self) -> str:
+        return f"{self.session.protocol}://{self.session.host}:{self.session.port}"
```
t1m0thyj, line 70: @shalearkane Base path should be appended to the end of the URL if one exists (`self.session.basePath or ''`).

tests/integration/test_zosmf.py
```diff
@@ -2,16 +2,16 @@
 from decouple import config
```
t1m0thyj, line 2: @richardgreg Please remove this import which is now unused.

Also, all the other integration test files need to have the same changes made in their setUp stage as this file (replacing the load of `decouple.config` and usage of `connection_dict` with the new ProfileManager class).

After updating all the integration tests, you can comment out the `ZOWE_TEST_PROFILE` in your .env file so the tests won't be able to find a v1 profile, and verify that all the tests pass using your v2 configuration (the zowe.config.json that we set up).

src/core/zowe/core_for_zowe_sdk/zosmf_profile2.py
shalearkane: Yes, its working on Linux ☺️

t1m0thyj:
Merging for now because other PRs depend on it, if there are any requested changes from future reviewers we can address them in a follow-up PR 🙂

gejohnston (approved):
I am not familiar with python, but the logic changes made sense.

Issue #58, opened 2022-06-07
[zos-files] Add method to create a zFS file system


t1m0thyj:
Add a method to the Python zos-files SDK to call the z/OSMF REST API `POST /zosmf/restfiles/mfs/zfs/<file-system-name>`

* z/OSMF REST API documentation: https://www.ibm.com/docs/en/zos/2.5.0?topic=interface-create-zos-unix-zfs-filesystem
* Python zos-files SDK in the "next" branch: https://github.com/zowe/zowe-client-python-sdk/tree/next/src/zos_files
* Node.js zos-files SDK for comparison: https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/create/Create.ts

aadityasinha-dotcom:
Can I work on this :slightly_smiling_face:

t1m0thyj:
@aadityasinha-dotcom I've responded on Slack. For now the zos-files issues are reserved. I created some new issues (#77 and #78) for other SDKs - feel free to take a look at those :slightly_smiling_face:

Issue #59, opened 2022-06-07
[zos-files] Add method to delete a zFS file system


t1m0thyj:
Add a method to the Python zos-files SDK to call the z/OSMF REST API `DELETE /zosmf/restfiles/mfs/zfs/<file-system-name>`

* z/OSMF REST API documentation: https://www.ibm.com/docs/en/zos/2.5.0?topic=interface-delete-zos-unix-zfs-filesystem
* Python zos-files SDK in the "next" branch: https://github.com/zowe/zowe-client-python-sdk/tree/next/src/zos_files
* Node.js zos-files SDK for comparison: https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/delete/Delete.ts

PR #103, opened 2022-08-01
Add methods to create and delete a zFS file system

richardgreg:
Fixes: #58 #59

Signed-off-by: Richard Gregory <richardgrecoson@gmail.com>


src/core/zowe/core_for_zowe_sdk/constants.py
```diff
@@ -14,6 +14,7 @@
     "SecureValuePrefix": "managed by ",
     "TsoSessionNotFound": "IZUG1126E",
     "ZoweCredentialKey": "Zowe-Plugin",
+    "MaxAllocationQuantity": 16777215,
```
t1m0thyj, line 17: Thanks for moving this constant into the z/OS Files SDK 👍 

Please remove this duplicate version of the constant from the Core SDK.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -10,8 +10,9 @@
 Copyright Contributors to the Zowe Project.
 """
 
-from zowe.core_for_zowe_sdk import SdkApi
+from zowe.core_for_zowe_sdk import SdkApi, constants
```
t1m0thyj, line 13: Please update the import to reference `constants` from the same SDK, not from the Core SDK.

PR #97, opened 2022-07-10
[zosmf] Added a method to list the systems defined to z/OSMF

aadityasinha-dotcom:
Added a method to `list the systems` defined to z/OSMF
Signed-off-by: Aaditya Sinha [aadityasinha009@gmail.com](mailto:aadityasinha009@gmail.com)

PR #106, opened 2022-08-11
[zos-jobs] Added support for `cancel_job` version number on requests

aadityasinha-dotcom:
Added an optional `modifyVersion` parameter to the `cancel_job` API with a default value of `"2.0"`


t1m0thyj:
@aadityasinha-dotcom Thanks for your contribution 🙂 Please add unit and integration tests for the `cancel_job` API modeled after the ones you wrote in #97.

src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
```diff
@@ -80,12 +80,10 @@ def cancel_job(self, jobname: str, jobid: str, modify_version="2.0"):
...
-        custom_args["json"] = {
-            "request": "cancel",
-            "version": modify_version
-        }
-
-        response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[202, 200])
+        custom_args["json"] = {"request": "cancel", "version": modifyVersion}
+        response_json = self.request_handler.perform_request("PUT", custom_args, expected_code = [202])
+        custom_args["json"] = {"request": "cancel"}
+        response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[202])
```
zFernand0, line 86: Please remove the duplicate REST calls and use `202, 200` for the `expected_code`.

zFernand0 (changes requested):
Thanks for working on this.
I left some comments that could improve the overall usage of the API 😋 

One additional consideration.
There are some z/OSMF REST APIs that don't return all headers if the response is empty.
For that reason, maybe we should consider the following small change too: 
From `self.response.headers['Content-Type']` to `self.response.headers.get('Content-Type')` in 
`src/core/zowe/core_for_zowe_sdk/request_handler.py`

(Thanks @t1m0thyj for helping me understand what was happening)

t1m0thyj:
Looks like this enhancement was already made in #124. Sorry about that 😢 Going forward, we can avoid this happening again by making sure to assign issues to avoid duplicate work 🙂

aadityasinha-dotcom:
> Looks like this enhancement was already made in #124. Sorry about that 😢 Going forward, we can avoid this happening again by making sure to assign issues to avoid duplicate work 🙂

No worries

Issue #79, opened 2022-06-15
Add readmes for each package on PyPI


t1m0thyj:
![image](https://user-images.githubusercontent.com/22344007/173843168-0fbc3171-535e-485b-aa59-df9d50d38e9c.png)

antondelchev:
Could I work on this one?

PR #109, opened 2022-08-15
Anton/contribute readmes

antondelchev:
Add readmes for each package on PyPI #79


t1m0thyj (commented):
Seems like I accidentally merged this PR without approving it somehow 😢 

Confirmed with @antondelchev that all these examples have been tested. Most worked but there were a few errors encountered. Will follow up on these in the linked issue.

PR #110, opened 2022-08-17
Anton/contribute create dataset

antondelchev:
Signed-off-by: Anton Delchev <dlchv.mail@gmail.com>


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -156,6 +158,10 @@ def create_data_set(self, dataset_name, options = {}):
         json
         """
 
+        if options.get("like") is None:
+            if options.get("primary") is None and options.get("lrecl") is None:
```
t1m0thyj, line 162: I believe this `and` should be an `or`, to throw an error when just one option is missing.

t1m0thyj (commented):
Thank you for adding the `create_default_data_set` method, I think it will be very useful 👍 

Left a comment 🙂 Could you please also update this PR with the "next" branch to resolve conflicts?

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
antondelchev: Thank you for the feedback Timothy, I'll make the necessary changes!

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -217,6 +223,90 @@ def create_data_set(self, dataset_name, options = {}):
...
+                "blksize": 27998,
+                "lrecl": 27998,
+                "dirblk": 25
+            }
+        elif default_type == "vsam":
+            options = {
+                "dsorg": "INDEXED",
+                "alcunit": "KB",
+                "primary": 840
+            }
```
t1m0thyj, line 301: VSAM is a special data set type that cannot be created with the same API as other types.

To create a VSAM, you need to use the "Access Method Services Interface" documented [here](https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-access-method-services). This is added scope outside of the issue addressed in this PR, so I'd suggest removing VSAM for now and we can handle it as a separate issue in the future 🙂

antondelchev: Alright, thank you, I'll edit this!

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -163,6 +165,10 @@ def create_data_set(self, dataset_name, options = {}):
         json
         """
 
+        if options.get("like") is None:
+            if options.get("primary") is None or options.get("lrecl") is None:
+                raise KeyError
+
         for opt in ("volser", "unit", "dsorg", "alcunit",
```
zFernand0, line 172: I believe that whenever the `like` option is specified in the `dict`, we shouldn't be adding any default values (e.g. `alcunit: TRK`, `recfm: F`)
Maybe we just need to include this `for` loop inside the `if('like') is None` condition 😋

zFernand0 (changes requested):
I agree that `create_default_data_set` is super useful! 🥳 
I do have one comment (and thanks @t1m0thyj for helping me understand what was happening)

I left a comment which I hope makes sense 😋

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
antondelchev: Thank you, I will change that!

t1m0thyj (approved):

t1m0thyj (changes requested):
Please fix the failing unit test 🙂

t1m0thyj (approved):

PR #113, opened 2022-08-19
Loading of Global and User Configs

shalearkane:
Now the Profile Manager class has been split into two parts :
1. Config File : This just concerns itself with just one file. It handles all the auto-discovering and loading of profiles (prepopulated form base profile properties or just the profile properties). It always raises exceptions when it faces any issues though when it loads secure configs, it suppresses exceptions.
2. ProfileManager : This class contains the logic to merge the different properties of profiles (from the Project Config and the Project User Config as well as the Global Config and Global User Config). This class handles all the exceptions raised in the Config File to provide a smooth user experience.

Config File is recommended when the user wants to load just one file while ProfileManager automates the merging of profiles.

TODO : Modify tests that checks if correct exceptions are being loaded.


src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -13,146 +13,198 @@
 import base64
 import os.path
 import re
+import sys
 import warnings
+from dataclasses import dataclass
 from typing import Union
-import sys
 
 import jsonc
```
t1m0thyj, line 21: The `jsonc` module doesn't seem to handle comments well. If you comment out a line in `zowe.config.json` by prefixing it with a double slash (`//`), then `jsonc` fails to load it.

I know we had a discussion about `jsonc` packages before and there didn't seem to be many options that are actively maintained. I quickly tested the `commentjson` package and although it was last updated in 2020, it seems to handle JSON files that contain comments correctly.

Could you replace the `jsonc` import here and the reference to it in "requirements.txt" with the `commentjson` package instead? I'm also open to other ideas if there are alternative packages you think would be a better fit 🙂

tests/unit/fixtures/zowe.config.user.json
```diff
@@ -0,0 +1,56 @@
+{
```
t1m0thyj, line 1: Please add some line and block comments in either this file or the `zowe.config.json` fixture. The contents of the comments don't matter, they'll just help to verify that ProfileManager is able to load JSON with comments.

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -13,146 +13,198 @@
...
+            self.autodiscover_config_dir()
 
-    @property
-    def config_filepath(self) -> Union[str, None]:
-        """Get the full Zowe z/OSMF Team Config filepath"""
-        return self._config_filepath
+        with open(self.filepath, encoding="UTF-8", mode="r") as fileobj:
+            profile_jsonc = jsonc.load(fileobj)
+
+        print(profile_jsonc)
```
t1m0thyj, line 101: Please remove this print statement.

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -299,19 +251,172 @@ def load_credentials(self) -> dict:
...
+        project_profile_name: Union[str, None] = None
+        global_profile_name: Union[str, None] = None
+
+        # get Project Profile
+        try:
+            project_profile, project_profile_name = self.project_config.get_profile(
+                profile_name=profile_name, profile_type=profile_type
+            )
+        except Exception as exc:
+            warnings.warn(f"Could not load Project Config {self.project_config.name}")
```
t1m0thyj, line 350: Thanks for replacing errors with warnings where the exception isn't fatal and it makes sense to try continuing 👍 

For debugging purposes it would be useful to include the exception message in the warning. Could you update this and all the "Could not load X Config" warnings below to include the exception? Something like this:
```
warnings.warn(f"Could not load Project Config {self.project_config.name}:\n{exc}")
```

t1m0thyj:
Closing in favor of #131

Issue #74, opened 2022-06-07
[core] Validate that zowe.config.json file matches schema


t1m0thyj:
Given a `zowe.config.json` file that references a `zowe.schema.json` file, validate that the contents of the JSON file conform to the schema.

PR #116, opened 2022-08-24
Anton/contribute validate json

antondelchev:
Validating that zowe.config.json matches zowe.schema.json. Related to #74


src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+
+    config_json = ""
+    schema_json = ""
+
+    with open(path_config_json) as f:
+        config_json = json.load(f)
+
+    with open(path_schema_json) as f:
+        schema_json = json.load(f)
```
t1m0thyj, line 40: To support comments in JSON files, we want to use the "commentjson" package.

Once #113 is merged, this will be added as a dependency to requirements.txt 🙂

src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -18,29 +18,25 @@ def validate_config_json(path_config_json: str, path_schema_json: str):
...
-        config_json = commentjson.load(f)
+    with open(path_config_json) as file:
+        config_json = commentjson.load(file)
 
-    with open(path_schema_json) as f:
-        schema_json = commentjson.load(f)
+    with open(path_schema_json) as file:
+        schema_json = commentjson.load(file)
 
     result = validate(instance=config_json, schema=schema_json)
```
boris-bc, line 40: Is `config_json` or `schema_jsopn`even  defined here?
I would propose this to be rewritten like this:
```
with open(path_config_json) as config_file, open(path_schema_json) as schema_file:
    config_json = commentjson.load(config_file)
    schema_json = commentjson.load(schema_file)
    return validate(instance=config_json, schema=schema_json)
```

boris-bc: You don't need to check `if result` if you anyway would return None `if not result`

tests/unit/test_zowe_core.py
```diff
@@ -35,6 +39,88 @@ def keyring_get_password(serviceName: str, username: str):
...
+                },
+                "base": {
+                    "description": "Default base profile",
+                    "type": "string"
+                }
+            }
+        },
+    }
+}
+
```
boris-bc, line 123: I would move those definitions closer to the method they are used in or even better, define them in separate files and load them from the tests with `json.open(open("path"))`

tests/unit/test_zowe_core.py
```diff
@@ -297,3 +383,47 @@ def test_secure_values_loading_exception(self, get_pass_func):
...
+        test_values = [
+            ((matching_config_json, matching_schema_json), True),
+            ((not_matching_config_json, not_matching_schema_json), False),
+        ]
+
+        current_dir = os.getcwd()
+
+        for test_case in test_values:
+            if test_case[1]:
+                with open("config.json", "w") as file:
```
boris-bc, line 402: Why do you need to create those files temporarily and then delete them? Can't you have a `resources` folder for your tests and add those files there permanently? Their content is not dinamyc, they are statically defined in the python script.

tests/unit/test_zowe_core.py
```diff
@@ -297,3 +383,47 @@ def test_secure_values_loading_exception(self, get_pass_func):
...
+                    commentjson.dump(matching_config_json, file)
+
+                with open("schema.json", "w") as file:
+                    commentjson.dump(matching_schema_json, file)
+
+                expected = validate(test_case[0][0], test_case[0][1])
+                result = validate_config_json(current_dir + "/config.json", current_dir + "/schema.json")
+
+                self.assertEqual(result, expected)
```
boris-bc, line 411: You should assert `expected` and `result` are not `None`.

tests/unit/test_zowe_core.py
```diff
@@ -297,3 +383,47 @@ def test_secure_values_loading_exception(self, get_pass_func):
             prof_manager = ProfileManager()
             prof_manager.config_dir = self.custom_dir
             props: dict = prof_manager.load("ssh")
+
+
+class TestValidateConfigJsonClass(unittest.TestCase):
+    """Testing the validate_config_json function"""
+
+    def test_validate_config_json_parameterized(self):
```
boris-bc, line 391: You only have two test cases, you handle both cases separately with an if statement and thus parametrization does not make sense. Rather have two separate test methods:
`test_validate_config_json_valid`
`test_validate_config_json_not_valid`

boris-bc (changes requested):

src/core/zowe/core_for_zowe_sdk/validators.py
antondelchev: A 'with' statement does not create a scope, for the 'if result' I'll just return what you suggested.

tests/unit/test_zowe_core.py
antondelchev: I'll refactor this as well.

tests/unit/test_zowe_core.py
antondelchev: Ok, I'll put them in a separate folder and import them from there.

tests/unit/test_zowe_core.py
antondelchev: Ok, I'll make two separate tests.

tests/unit/test_zowe_core.py
```diff
@@ -387,43 +304,32 @@ def test_secure_values_loading_exception(self, get_pass_func):
...
+        path_to_invalid_schema = FIXTURES_PATH + "/invalid.zowe.schema.json"
 
-                with self.assertRaises(Exception) as expected_info:
-                    validate(test_case[0][0], test_case[0][1])
+        invalid_config_json = json.load(open(path_to_invalid_config))
+        invalid_schema_json = json.load(open(path_to_invalid_schema))
 
-                with self.assertRaises(Exception) as actual_info:
-                    validate_config_json(current_dir + "/config.json", current_dir + "/schema.json")
+        with self.assertRaises(Exception) as expected_info:
```
boris-bc, line 329: Shouldn't you expect here `ValidationError`?

tests/unit/test_zowe_core.py
```diff
@@ -387,43 +304,32 @@ def test_secure_values_loading_exception(self, get_pass_func):
...
+            validate(invalid_config_json, invalid_schema_json)
 
-                self.assertEqual(str(actual_info.exception), str(expected_info.exception))
+        with self.assertRaises(Exception) as actual_info:
```
boris-bc, line 332: Shouldn't you expect here `ValidationError`?

tests/unit/test_zowe_core.py
antondelchev: Yes, I'll fix that.

tests/unit/test_zowe_core.py
antondelchev: I'll fix that as well.

boris-bc (approved):

boris-bc:
@antondelchev , the build is failing on unit tests. Please review and fix!

PR #123, opened 2022-08-30
Anton/contribute delete migrated

antondelchev:
Added method to delete a migrated data set. Related to #67


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -485,3 +486,43 @@ def list_unix_file_systems(self, file_path_name=None, file_system_name=None):
...
+            "wait": json.dumps(False),
+            "purge": json.dumps(False),
+        }
+
+        if wait:
+            data["wait"] = json.dumps(True)
+        
+        if purge:
+            data["purge"]: json.dumps(True)
+
```
skorzhevsky, line 522: Maybe just:
    data = {
            "request": "hdelete",
            "wait": json.dumps(wait),
            "purge": json.dumps(purge),
        }

antondelchev: Thank you!

t1m0thyj (approved):

boris-bc (approved):

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -567,6 +567,39 @@ def list_unix_file_systems(self, file_path_name=None, file_system_name=None):
         response_json = self.request_handler.perform_request("GET", custom_args, expected_code=[200])
         return response_json
 
+    def delete_migrated_data_set(self, dataset_name: str, wait=False, purge=False):
```
boris-bc, line 570: I think the order of those arguments should be swapped, first `purge` and then `wait` as last. If you do this change, review your parametrized test that it is still structured correctly.

boris-bc (changes requested):

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
antondelchev: Ok, will swap them and refactor the tests.

boris-bc (approved):

Issue #90, opened 2022-06-17
[zos-jobs] `cancel_job` should support version number on requests


t1m0thyj:
According to [IBM docs](https://www.ibm.com/docs/en/zos/2.5.0?topic=interface-cancel-job):
> To request asynchronous processing for this service, set the "version" property to 1.0. To request synchronous processing (the default), set "version" to 2.0 or omit the property from the request. If so, the system will attempt to process the request synchronously, if such processing is supported on the target JES subsystem.

To match the delete API (see #91), we should add an optional `modifyVersion` parameter to the `cancel_job` API with a default value of `"2.0"` (synchronous). Users can set it to `"1.0"` if they prefer asynchronous processing.

The value of the `modifyVersion` parameter should be passed on to the REST request in the `version` property of the payload.

PR #124, opened 2022-08-30
Added modify_version to cancel_job method

antondelchev:
Added modify_version optional parameter to the cancel_job method. Related to #90

### DIFF
--- a/src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
+++ b/src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
@@ -74,7 +74,7 @@ class Jobs(SdkApi):
             A JSON containing the result of the request execution
         """
         if modify_version not in ("1.0", "2.0"):
-            raise ValueError('Accepted values: "1.0" or "2.0".')
+            raise ValueError('Accepted values for modify_version: "1.0" or "2.0"')
 
         custom_args = self._create_custom_request_arguments()
         job_url = "{}/{}".format(jobname, jobid)
@@ -82,11 +82,8 @@ class Jobs(SdkApi):
         custom_args["url"] = request_url
         custom_args["json"] = {
             "request": "cancel",
-            "version": "2.0"
+            "version": modify_version
         }
-
-        if modify_version == "1.0":
-            custom_args["json"]["version"] = "1.0"
 
         response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[202, 200])
         return response_json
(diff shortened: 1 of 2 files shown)


src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
```diff
@@ -65,6 +65,8 @@ def cancel_job(self, jobname, jobid):
             The name of the job
         jobid: str
             The job id on JES
+        modify_version: str
+            Default ("2.0") specifies that the request is to be processed synchronously. For asynchronous processing - change the value to "1.0"
```
skorzhevsky, line 70: **IMHO**, it would be better to use a meaningful name, like "async_req=True", instead of "modify_version=..".
Also, as far as i can see https://www.ibm.com/docs/en/zos/2.4.0?topic=interface-cancel-job-purge-its-output, "default" value is "async", and sync works only for JES2. 
It is just my opinion. Thank you.

antondelchev: Alright, thank you, might make some edits here as well! Read the docs too, but I was just following the request from the opened issue:
"... we should add an optional modifyVersion parameter to the cancel_job API with a default value of "2.0" (synchronous). Users can set it to "1.0" if they prefer asynchronous processing."

boris-bc: I would recommend to stick to the description of the issue and interpret it. There was probably some thought behind the decision to request a `modify_version` parameter over a different one. If you think the design is not right, this should be discussed on a call with the squad and if a different design is considered, the issue should be updated or a new one should be created.

src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
```diff
@@ -73,6 +73,8 @@ def cancel_job(self, jobname, jobid, modify_version="2.0"):
         response_json
             A JSON containing the result of the request execution
         """
+        if modify_version not in ("1.0", "2.0"):
+            raise ValueError('Accepted values: "1.0" or "2.0".')
```
boris-bc, line 77: Extend the message to: "Accepted values for modify_version: "1.0" or "2.0""

boris-bc (changes requested):

tests/unit/test_zos_jobs.py
```diff
@@ -1,15 +1,15 @@
...
 
-class TestJobsClass(unittest.TestCase):
+class TestJobsClass(TestCase):
     """Jobs class unit tests."""
 
     def setUp(self):
         """Setup fixtures for Jobs class."""
-        self.connection_dict = {"host": "https://mock-url.com",
+        self.test_profile = {"host": "https://mock-url.com",
                                 "user": "Username",
```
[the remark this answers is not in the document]
antondelchev: I'll fix it.

PR #129, opened 2022-09-02
Only publish dev releases for labelled PRs

t1m0thyj:
Initially I set up the Release workflow to run on all pushes to the "next" branch that modify the "src" folder. The creation of a new GitHub release every time a PR is merged into the "next" branch may be undesirable. I do not want to spam our release history 🙂 

In this PR, the behavior is changed to only publish dev releases for PRs merged into the "next" branch with the "release-dev" label.

### DIFF
--- a/.github/workflows/dev-release.yaml
+++ b/.github/workflows/dev-release.yaml
@@ -1,19 +1,21 @@
 name: Zowe SDK Release
 
 on:
-  push:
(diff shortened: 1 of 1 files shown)


zFernand0 (approved):
LGTM! 😋 
seems like a very simple and effective change 👍

Issue #71, opened 2022-06-07
[core] Merge global config with project config


t1m0thyj:
Given `zowe.config.json` and `zowe.config.user.json` files in home directory (global config) and current directory (project config), merge their profile properties together. Project config should take precedence, and we should search up the directory tree for it.

Issue #70, opened 2022-06-07
[core] Load secure profile properties from vault


t1m0thyj:
Given `zowe.config.json` and `zowe.config.user.json` files, load their secure properties from the vault (`Zowe/secure_config_props`).

Issue #69, opened 2022-06-07
[core] Load profiles from zowe.config.user.json file


t1m0thyj:
Given a `zowe.config.user.json` file, read profile properties from it and merge them into profiles loaded from `zowe.config.json`.

Issue #68, opened 2022-06-07
[core] Load profiles from zowe.config.json file


t1m0thyj:
Given a `zowe.config.json` file, read v2 profiles from it to replace the [ZosmfProfile](https://github.com/zowe/zowe-client-python-sdk/blob/next/src/core/zowe/core_for_zowe_sdk/zosmf_profile.py) API that loads v1 profiles. Must use a JSON loader that supports comments (JSONC).

PR #131, opened 2022-09-06
Refactored Profile Manager Class

shalearkane:
Now the Profile Manager class has been split into two parts :

- Config File : This just concerns itself with just one file. It handles all the auto-discovering and loading of profiles (prepopulated form base profile properties or just the profile properties). It always raises exceptions when it faces any.

- ProfileManager : This class contains the logic to merge the different properties of profiles (from the Project Config and the Project User Config as well as the Global Config and Global User Config). This class handles all the exceptions raised in the Config File to provide a smooth user experience.

Config File is recommended when the user wants to load just one file while ProfileManager automates the merging of profiles.

This is the duplicate of #113 since I had to rebase against the next branch

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -125,12 +125,16 @@ class ConfigFile:
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
     ) -> Tuple[dict, str]:
+        """
+        Load given profile with values populated from base profile
+        """
         if self.profiles is None:
             self.init_from_file()
 
         if profile_name is None and profile_type is None:
             raise ProfileNotFound(
-                "Could not find profile as both profile_name and profile_type is not set."
+                profile_name=profile_name,
+                error_msg="Could not find profile as both profile_name and profile_type is not set.",
             )
 
         if profile_name is None:
@@ -194,7 +198,7 @@ class ConfigFile:
             profilename = self.defaults[profile_type]
         except KeyError:
             warnings.warn(
-                "Given profile type has no default profilename",
+                f"Given profile type '{profile_type}' has no default profilename",
                 ProfileParsingWarning,
             )
         else:
@@ -208,14 +212,14 @@ class ConfigFile:
                     return key
             except KeyError:
                 warnings.warn(
-                    f"Profile {key} has no type attribute",
+                    f"Profile '{key}' has no type attribute",
                     ProfileParsingWarning,
                 )
 
         # if no profile with matching type found, we raise an exception
         raise ProfileNotFound(
             profile_name=profile_type,
-            error_msg=f"No profile with matching profile_type {profile_type} found",
+            error_msg=f"No profile with matching profile_type '{profile_type}' found",
         )
 
     def load_profile_properties(self, profile_name: str) -> dict:
@@ -226,7 +230,7 @@ class ConfigFile:
         try:
             props = self.profiles[profile_name]["properties"]
         except Exception as exc:
-            raise ProfileNotFound("Profile {profile_name} not found", error_msg=exc)
+            raise ProfileNotFound("Profile {profile_name} not found", error_msg=exc) from exc
 
         secure_fields: list = self.profiles[profile_name].get("secure", [])
 
@@ -293,7 +297,7 @@ class ConfigFile:
                 raise SecureProfileLoadFailed(
                     constants["ZoweServiceName"],
                     error_msg="No credentials found for loaded config file as well as for global config",
-                )
+                ) from exc
             else:
                 warnings.warn(
                     f"Credentials not found for given config, using global credentials {GLOBAL_CONFIG_PATH}",
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -16,19 +16,17 @@ import re
 import sys
 import warnings
 from dataclasses import dataclass, field
-from typing import List, Optional, NamedTuple
+from typing import Optional, NamedTuple
 
 import commentjson
 
 from .constants import constants
 from .custom_warnings import (
-    ProfileNotFoundWarning,
     ProfileParsingWarning,
     SecurePropsNotFoundWarning,
 )
-from .exceptions import ProfileNotFound, SecureProfileLoadFailed, SecureValuesNotFound
+from .exceptions import ProfileNotFound, SecureProfileLoadFailed
 from .profile_constants import (
-    BASE_PROFILE,
     GLOBAL_CONFIG_NAME,
     TEAM_CONFIG,
     USER_CONFIG,
@@ -52,7 +50,7 @@ CURRENT_DIR = os.getcwd()
 class Profile(NamedTuple):
     data: dict = {}
     name: str = ""
-    secure_props_not_found: list = []
+    missing_secure_props: list = []
 
 
 @dataclass
@@ -79,7 +77,7 @@ class ConfigFile:
     profiles: Optional[dict] = None
     defaults: Optional[dict] = None
     secure_props: Optional[dict] = None
-    _secure_props_not_found: list = field(default_factory=list)
+    _missing_secure_props: list = field(default_factory=list)
 
     @property
     def filename(self) -> str:
@@ -134,9 +132,11 @@ class ConfigFile:
         profile_type: Optional[str] = None,
     ) -> Profile:
         """
-        Load given profile with values populated from base profile
-
-        Returns a namedtuple called Profile
+        Load given profile including secure properties and excluding values from base profile
+        Returns
+        -------
+        Profile
+            Returns a namedtuple called Profile
         """
         if self.profiles is None:
             self.init_from_file()
@@ -154,7 +154,7 @@ class ConfigFile:
 
         props: dict = self.load_profile_properties(profile_name=profile_name)
 
-        return Profile(props, profile_name, self._secure_props_not_found)
+        return Profile(props, profile_name, self._missing_secure_props)
 
     def autodiscover_config_dir(self) -> None:
         """
@@ -164,8 +164,8 @@ class ConfigFile:
         -------
         None
 
-        Return path if it finds the config directory,
-        Else, it returns None
+        Sets path if it finds the config directory,
+        Else, it raises an Exception
         """
 
         current_dir = CURRENT_DIR
@@ -192,8 +192,7 @@ class ConfigFile:
         -------
         str
 
-        Return exact profilename of the profile to load from the mentioned type
-
+        Return the exact profilename of the profile to load from the mentioned type.
         First tries to look into the defaults, if not found,
         then it tries to iterate through the profiles
         """
@@ -228,7 +227,7 @@ class ConfigFile:
 
     def load_profile_properties(self, profile_name: str) -> dict:
         """
-        Load profile properties including secure properties
+        Load profile properties given profile_name including secure properties
         Returns
         -------
         dictionary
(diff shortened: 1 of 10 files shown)


t1m0thyj (commented):
@shalearkane Thanks for updating #113 👍 

Please address this comment, I believe all the others have been addressed:
> Please add some line and block comments in either the `zowe.config.json` or `zowe.config.user.json` fixture. The contents of the comments don't matter, they'll just help to verify that ProfileManager is able to load JSON with comments.

t1m0thyj (changes requested):
Thanks for addressing feedback on this PR 🙂 

After this PR is merged, I think we can close out several issues: #68, #69, #70, #71
I've opened some issues for potential enhancements that can be addressed in the future: #134, #135, #136

Before merging this PR, I'm having some difficulties with testing it. These are my files:
**zowe.config.json**
```json
{
    "$schema": "./zowe.schema.json",
    "profiles": {
        "zosmf": {
            "type": "zosmf",
            "properties": {
                "port": 443
            },
            "secure": []
        },
        "base": {
            "type": "base",
            "properties": {
                "host": "example.com",
                "rejectUnauthorized": false
            },
            "secure": [
                "user",
                "password"
            ]
        }
    },
    "defaults": {
        "zosmf": "zosmf",
        "base": "base"
    },
    "autoStore": true
}
```
**zowe.config.user.json**
```json
{
    "$schema": "./zowe.schema.json",
    "profiles": {
        "base": {
            "type": "base",
            "properties": {
                "user": "XXX",
                "password": "XXX"
            },
            "secure": []
        }
    }
}
```
**test.py**
```python
from zowe.core_for_zowe_sdk import ProfileManager
profile = ProfileManager().load(profile_type="zosmf")
print(profile)
```

An empty object is printed to the console. I would expect it to load the host, port, user, password, and rejectUnauthorized properties.

shalearkane:
I will try to find out what the exact issue is but for future reference,

I wrote ConfigFile class and ProfileManager class with two behaviour :

1. ConfigFile class with raise Exceptions all the time, so even if everything goes right but it doesn't find secure credentials, it will throw an Exception. The sideeffect is that even if the secure fields in the json is empty, it will try to load the secure props from keyring and if it doesn't, it will throw up an error and say that it could not find the properties, hence all it will return is an empty dictionary. 
2. ProfileManager class will convert all those Exception into respective warnings.

So the issue I was thinking that on profiles which doesn't need to fetch secure props, this try to load global_secure_props and if that fails too, this will raise and exception and nothing will be returned. 

This behavior arises from the fact that ConfigFile tries to load secure properties before looking at the json as a part of the init_from_file() hence if that fails, nothing is returned

``` python
def init_from_file(self) -> None:
        if self.filepath is None:
            self.autodiscover_config_dir()

        with open(self.filepath, encoding="UTF-8", mode="r") as fileobj:
            profile_jsonc = commentjson.load(fileobj)

        self.profiles = profile_jsonc["profiles"]
        self.defaults = profile_jsonc["defaults"]

        self.load_secure_props()
```

So I have two approaches here :
1. Convert the Exceptions in `load_secure_props(self)` function to warnings, or
2. Call the function `load_secure_props(self)` only if we have secure properties to load

shalearkane:
This is failing because zowe.config.user.json has no profile named "zosmf", I am updating the code to return the base profile when this happens.

t1m0thyj:
> This is failing because zowe.config.user.json has no profile named `zosmf`, should I update the code to return the base profile when no profile is found?

Yes, a "zosmf" profile should not be required in zowe.config.json or zowe.config.user.json. When `profile_type="zosmf"` is requested and no "zosmf" profile exists, we want to fall back to loading properties from a "base" profile. It would be ok to show a warning in this case 🙂

shalearkane:
Yes, I fixed the issue :slightly_smiling_face: . Now it should be loading the config correctly.

t1m0thyj:
> Yes, I fixed the issue 🙂 . Now it should be loading the config correctly.

Thanks 🙂 Now it is loading the user config correctly, but failing to merge in properties from the project config. The output I am seeing from my sample script above is:
```
{'user': 'XXX', 'password': 'XXX'}
```

The ProfileManager should also be loading "port" from the zosmf profile in zowe.config.json and "host"/"rejectUnauthorized" from the base profile in zowe.config.json and merging these properties in.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -0,0 +1,301 @@
...
+        First tries to look into the defaults, if not found,
+        then it tries to iterate through the profiles
+        """
+        # try to get the profilename from defaults
+        try:
+            profilename = self.defaults[profile_type]
+        except KeyError:
+            warnings.warn(
+                "Given profile type has no default profilename",
+                ProfileParsingWarning,
```
zFernand0, line 198: Should we print the given profile type in the warning ?
`Given profile type 'base' has no default profilename"`
It may be helpful 😋

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -0,0 +1,301 @@
...
+            except KeyError as exc:
+                raise SecureProfileLoadFailed(
+                    constants["ZoweServiceName"],
+                    error_msg="No credentials found for loaded config file as well as for global config",
+                )
+            else:
+                warnings.warn(
+                    f"Credentials not found for given config, using global credentials {GLOBAL_CONFIG_PATH}",
+                    SecurePropsNotFoundWarning,
+                )
```
zFernand0, line 301: I'm probably missing something, but why do we throw an error when credentials are found in the `global_config_path` ?

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -58,260 +71,169 @@ def config_appname(self) -> str:
...
+                    f"Secure properties of profile type '{profile_type}' from file '{cfg.filename}' were not found "
+                    f"hence profile not loaded.",
+                    SecurePropsNotFoundWarning,
+                )
+        except Exception as exc:
+            warnings.warn(
+                f"Could not load {config_type} '{cfg.filename}' at '{cfg.filepath}'"
+                f"because {type(exc).__name__}'{exc}'.",
+                ConfigNotFoundWarning,
+            )
```
zFernand0, line 159: I think we let `cfg.get_profile` raise the `ProfileNotFound` error.
I'm wondering if all `__init__`  in `exceptions.py` should have default values (in this case `profile_name` and `error_msg`) in order to print a better error message

current behavior
```
TypeError'ProfileNotFound.__init__() missing 1 required positional argument: 'error_msg''."
```

with `""` as defaults
```
ProfileNotFound: Failed to load profile Could not find profile as both profile_name and profile_type is not set.: 
```

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -58,260 +71,169 @@ def config_appname(self) -> str:
...
-        """
-        base_props: dict = {}
-
-        try:
-            base_profile_name = self.get_profilename_from_profiletype(
-                profile_jsonc=profile_jsonc, profile_type="base"
+        except ProfileNotFound:
+            warnings.warn(
+                f"Profile not found in file '{cfg.filename}', trying to return base profile instead.",
+                ProfileNotFoundWarning,
```
zFernand0, line 124: - Is the warning here required if the base profile exists ?
- Is there a way to turn off all warnings ?

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -58,260 +71,169 @@ def config_appname(self) -> str:
...
+        )
+        global_base_profile.update(global_base_user_profile)
+
+        # get Global Profile
+        global_profile, global_profile_name = self.get_profile(
+            self.global_config,
+            profile_name=None,
+            profile_type="base",
+            config_type="Global Config",
+        )
```
zFernand0, line 220: I believe the intention was to be able to load service profiles from the global config, but `base` was hardcoded 😢 
I tried loading a service profile from my global layers and none of them were loaded, only the base profile properties showed up 😋

zFernand0 (changes requested):
Most of my comments are just strange questions.
I do have one requested change and it's around loading service profiles from the global layers 😋

src/core/zowe/core_for_zowe_sdk/profile_manager.py
shalearkane: Thanks for finding the bug, I missed this while refactoring the code 😅 ... I will double check every other thing and address all of the issues

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -58,260 +71,169 @@ def config_appname(self) -> str:
...
+        return self.project_config.filepath
 
-            if os.path.isfile(path):
-                self._config_dir = current_dir
+    @staticmethod
+    def get_profile(
+        cfg: ConfigFile,
+        profile_name: Union[str, None],
+        profile_type: Union[str, None],
+        config_type: str,
```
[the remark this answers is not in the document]
shalearkane: We cannot use `str | None` because this syntax was introduced only in Python 3.10 ... however, yes, Optional is a better choice, I will update my code

src/core/zowe/core_for_zowe_sdk/config_file.py
shalearkane: We are just warning that credentials specific to that config path is not found ... because the credentials stored under global_config_path may not be appropriate to load.

src/core/zowe/core_for_zowe_sdk/profile_manager.py
t1m0thyj: Regarding if there a way to turn off all warnings, I believe there is 🙂 https://stackoverflow.com/questions/14463277/how-to-disable-python-warnings

Would you prefer for the warning here to be removed in the case that a base profile exists to fall back to?

t1m0thyj: Discussed with @shalearkane and agreed it would be nice to have a `show_warnings` argument on the ProfileManager constructor which defaults to `true`, and if set to `false` will disable all the warnings.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -0,0 +1,301 @@
...
+                if re.match(
+                    "profiles\\." + profile_name + "\\.properties\\.[a-z]+", key
+                ):
+                    property_name = key.split(".")[3]
+                    if property_name in secure_fields:
+                        props[property_name] = value
+                        secure_fields.remove(property_name)
+
+            if len(secure_fields) > 0:
+                raise SecureValuesNotFound(secure_fields)
```
t1m0thyj, line 248: Instead of raising this error at the config file level, can we delay throwing a `SecureValuesNotFound` error until we have loaded all the config files in `ProfileManager.load`?

It should be allowable to store "user" and "password" in a zowe.config.user.json file instead of in the secure vault 🙂

t1m0thyj: After the above comment is addressed, we can verify that the `SecureValuesNotFound` error is issued when secure credentials are missing.

t1m0thyj: @shalearkane I updated some of the logic for merging configs and profiles to simplify it a bit 🙂 I believe the comment above still needs to be addressed. Please let me know if you have any questions.

t1m0thyj (approved):

zFernand0 (approved):

t1m0thyj:
@shalearkane Thanks for updating tests, ready to merge 🙂

PR #144, opened 2023-01-09
Added support for Nested team config

### LINKED ISSUE #135: Support nested team config files
Currently only top-level profiles get loaded. We should support a zosmf child profile with a typeless parent profile like the following:
```
{
    "$schema": "./zowe.schema.json",
    "profiles": {
        "lpar1": {
            "profiles": {
                "zosmf": {
                    "type": "zosmf",
                    "properties": {
                        "port": 443
                    },
                    "secure": []
                }
            },
            "properties": {
                "host": "example.com",
                "rejectUnauthorized": true
            },
            "secure": [
                "user",
                "password"
            ]
        }
    },
    "defaults": {
        "zosmf": "lpar1.zosmf"
    },
    "autoStore": true
}
```

aadityasinha-dotcom:
Resolves #135 
Signed-off-by: aadityasinha-dotcom <aadityasinha009@gmail.com>


src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -238,7 +238,11 @@ def load_profile_properties(self, profile_name: str) -> dict:
         from the profile dict and populate fields from the secure credentials storage
         """
         try:
-            props = self.profiles[profile_name]["properties"]
+            props = self.profiles
+            lst = profile_name.split(".")
+            for i in range(len(lst)-1):
+                props = props[lst[i]]["profiles"]
+            props = props[lst[len(lst)-1]]["properties"]
```
zFernand0, line 245: I still need to do some testing manually, but will this implementation retain the  `A.properties` or always go for the `LAST.properties` (in a structure like `A.B.C...LAST`)?

zFernand0 (commented):
I believe this is going in the right direction.
Can we have some tests added just to verify tht everything works as expected?

src/core/zowe/core_for_zowe_sdk/config_file.py
aadityasinha-dotcom: It will give something like this

tests/unit/test_zowe_core.py
```diff
@@ -258,6 +258,37 @@ def test_custom_file_and_custom_profile_loading(self, get_pass_func):
...
+
+        self.setUpCreds(custom_file_path, {
+            "profiles.zosmf.properties.user": "user",
+            "profiles.zosmf.properties.password": "password",
+        })
+
+        # Test
+        prof_manager = ProfileManager(appname=self.custom_appname)
+        prof_manager.config_dir = self.custom_dir
+        props: dict = prof_manager.load(profile_name="lpar1.zosmf")
```
aadityasinha-dotcom, line 280: It is giving error in this line. Any suggestions?

t1m0thyj: @aadityasinha-dotcom I approved the workflow and it seems that now this line is working but there are 2 tests failing:
```
FAILED tests/unit/test_zowe_core.py::TestZosmfProfileManager::test_custom_file_and_custom_profile_loading
FAILED tests/unit/test_zowe_core.py::TestZosmfProfileManager::test_profile_loading_with_user_overriden_properties
```

Could you please investigate these failing tests? Happy to help if you have any questions 🙂

t1m0thyj (changes requested):
Waiting on tests to be updated to use a nested zowe.config.json 🙂

tests/unit/test_zowe_core.py
```diff
@@ -172,6 +172,7 @@ def setUp(self):
         )
         self.fs.add_real_file(self.original_file_path)
         self.fs.add_real_file(self.original_user_file_path)
+        self.fs.add_real_file(os.path.join(FIXTURES_PATH, "nested.zowe.config.json"))
```
aadityasinha-dotcom, line 175: @t1m0thyj did the changes like you suggested :relaxed:

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -238,7 +238,11 @@ def load_profile_properties(self, profile_name: str) -> dict:
         from the profile dict and populate fields from the secure credentials storage
         """
         try:
-            props = self.profiles[profile_name]["properties"]
+            props = self.profiles
```
aadityasinha-dotcom, line 241: While passing `lpar1.zosmf` the `load_profile_properties` is getting `my_base` as profile_name, but it works fine `zosmf` where profile_name is `zosmf`.

aadityasinha-dotcom: hey @t1m0thyj did some changes. Can you test that?

aadityasinha-dotcom: @t1m0thyj here is the [team config files](https://github.com/zowe/zowe-client-python-sdk/files/11011719/config.zip) that I have tested for

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -237,14 +273,29 @@ def load_profile_properties(self, profile_name: str) -> dict:
...
+                warnings.warn(
+                        f"Profile {profile_name} not found",
+                        ProfileNotFoundWarning
+                        )
+            lst.pop()
+
+        secure_fields: list = []
+        lst = profile_name.split(".")
+        while len(lst) > 0:
+            secure_fields.extend(self.find_profile(".".join(lst), self.profiles).get("secure", []))
```
aadityasinha-dotcom, line 297: In the unit test, this line is `Attribute_error` for NoneType

src/core/zowe/core_for_zowe_sdk/profile_manager.py
t1m0thyj, line 1: Tested this PR with the sample config and profiles that we've been using. For the `my_zosmf` profile it is failing to load user and password:
Python SDK: `{'host': 'example.com', 'rejectUnauthorized': True, 'port': 443}`
Node.js SDK: `{"port":443,"host":"example.com","rejectUnauthorized":true,"user":"admin","password":"password"}`

The `ProfileManager.load` method loops through config files in the following order:
1. Project User Config (`./zowe.config.user.json`)
2. Project Config (`./zowe.config.json`)
3. Global User Config (`~/zowe.config.user.json`)
4. Global Config (`~/zowe.config.json`)

In the current implementation, profile properties are first loaded from each layer of the config individually, and merged together afterwards.

For the case of our sample config:
* The `user` and `password` properties are stored in the `my_base` profile of Project User Config (`./zowe.config.user.json`).
* This `my_base` profile is defined as the default base profile in Project Config (`./zowe.config.json`).
* At the time when we load Project User Config, we do not yet know that `my_base` is the default base profile, so the current logic will skip loading its properties. 😢 

To fix this, I believe refactoring `ProfileManager.load` will be necessary. We should first merge all the config layers together, and then load profile properties as a single operation from the resulting merged object.

For a detailed explanation of how config layers should be merged together, see [Zowe docs](https://docs.zowe.org/stable/user-guide/cli-using-understand-profiles-configs).

t1m0thyj (changes requested):

t1m0thyj (approved):
Approving per the above comment 😋

Issue #154, opened 2023-05-16
Document the project setup steps


bhavukkalra:
**Is your feature or enhancement request related to a problem or limitation? Please describe**

When a new contributor visits the project for contribution. It could be quiet overwhelming to see the whole project structure and different files that are contained in it!
Currently there are step by step docs available on how to setup the project locally.

**Describe your enhancement idea**

- Add the step by step commands required to setup the project
- Add a starter script "template" to confirm if the project was set up correctly
- Document the pitfalls that were faced in the issue #153 for python3.11

**Describe alternatives you've considered**

As this is a growing project, I am hoping the starter docs would change over time. Giving the new contributors something to get started with the setup and start contributing

**Provide any additional context**

I was thinking, I'll first start with the `readme.md` file and then gradually add this to our official docs at `/docs`. But that is open to discussion.

cc- @adam-wolfe @t1m0thyj

achiverram28:
Yeah , I can support for the macOS setup, because even I faced some issues in setting up . I can document the steps and help other macOs users to follow the steps , if they face any error

bhavukkalra:
That would be great actually. Could you point out specifically where you encountered the error with a UNIX like environment? 
I can go ahead and add those changes as well in the draft PR #155.

For me it was specifically the python3.11 version mismatch with the `lxml` package (`but that was not a windows based error`)

Did you encountered any errors specific to MACOS?

achiverram28:
In macOs , also , there was an issue because of lxml I was using python3 . For mac users they can downgrade their python versions using pyenv locally for their project and then run the pip install - r requirements.txt . After this all the imports work propely

PR #159, opened 2023-05-18
DOC: Add local setup steps

bhavukkalra:
**What It Does**
Adds started contribution docs to the project and closes #154 

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] added/updated automated tests
- [x] updated the changelog
- [x] followed the [contribution guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md)

**Additional Comments**
Since these are DOC changes the tests are not written along with this PR

Signed-off-by: bhavukkalra <bhavukkalra1786@gmail.com>

cc - @aadityasinha-dotcom

PR #160, opened 2023-05-18
DOC: Add local setup steps for macOS

achiverram28:
**What It Does**

Document the project setup steps  for the macOS setup and adds more feature for #154

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,29 @@ result = my_zosmf.get_info()
 ```
 The result will be a JSON object containing z/OSMF information
 
+## Local setup for macOs
+
+Make sure that the current python version is `3.10` or older
+
+If it is in an older version , you can downgrade to a lower versions by following the below steps:
+
+```bash
+brew install pyenv
+echo 'PATH=$(pyenv root)/shims:$PATH' >> ~/.zshrc
+pyenv install 3.10 ## Example for python 3.10 version
+```
+
+After this , you can clone the repository
+
+```
+git clone https://github.com/zowe/zowe-client-python-sdk.git
+pyenv local <version> ## you can get thsi by running pyenv versions
+## This will create a python-version file , this version will only be used locally for this project
+python -m venv <envname>
+source <envname>/bin/activate
+cd <project>
+pip install -r requirements.txt
+```
 
 # Acknowledgments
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -187,6 +187,67 @@ result = my_zosmf.get_info()
...
+Make sure that the current python version is `3.10` or older
+
+If it is in an older version , you can downgrade to a lower versions by following the below steps:
+
+```bash
+brew install pyenv
+
+echo 'PATH=$(pyenv root)/shims:$PATH' >> ~/.zshrc
+
+pyenv install 3.10 ## Example for python 3.10 version
```
traeok, line 201: `brew` may not be installed for some users, but it isn't listed as a pre-requisite for the added steps.

README.md
```diff
@@ -187,6 +187,67 @@ result = my_zosmf.get_info()
 ```
 The result will be a JSON object containing z/OSMF information
 
+## Local setup for macOs
+
+Make sure that the current python version is `3.10` or older
+
+If it is in an older version , you can downgrade to a lower versions by following the below steps:
```
traeok, line 194: Should this be something like "If your Python version is newer than 3.10, you can downgrade to a lower version by using a tool such as pyenv."

I think since the usage of `pyenv` isn't mandatory here and is separate from these setup steps, it may not apply to all users reading this. In this case, I would recommend mentioning `pyenv` as an option, but without going into detail on how to use it.

README.md
```diff
@@ -187,6 +187,67 @@ result = my_zosmf.get_info()
...
+
+## This will create a python-version file , this version will only be used locally for this project
+
+```
+
+Create a virtual environment
+
+```
+
+python -m venv <envname>
```
traeok, line 238: At the beginning of these steps, it is noted that "the current python version is `3.10` or older"..

However, `venv` will not work on Python 3.1 and 3.2. Maybe we should consider mentioning a minimum Python3 version, or add a note saying that `venv` is only supported for Python 3.3 and above.

traeok (changes requested):
Thanks for working on the Python SDK's documentation - I left a few comments. 

In addition, this seems pretty similar to #159 - maybe it can be consolidated into one PR?

t1m0thyj:
@achiverram28 Thanks for contributing to the Zowe Python SDK!

Since this PR has been inactive for a few months with unaddressed comments, closing it in favor of #204.

If you free up in the future and are interested in contributing, feel free to tackle any issues labeled "good-first-issue" 🙂

PR #178, opened 2023-05-30
pass in timeout

savannah-portilla1:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/sdk_api.py
+++ b/src/core/zowe/core_for_zowe_sdk/sdk_api.py
@@ -21,7 +21,7 @@ class SdkApi:
     Abstract class used to represent the base SDK API.
     """
 
-    def __init__(self, profile, default_url):  
+    def __init__(self, profile, default_url, timeout=30):  
         self.profile = profile
         session = Session(profile)
         self.session: ISession = session.load()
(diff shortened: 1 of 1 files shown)


src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -21,7 +21,7 @@ class SdkApi:
     Abstract class used to represent the base SDK API.
     """
 
-    def __init__(self, profile, default_url):  
+    def __init__(self, profile, default_url, timeout=30):
```
t1m0thyj, line 24: Could this be generalized to `**session_args` so that all session arguments (including but not specific to timeout) can be overridden?
```python
    def __init__(self, profile, default_url, **session_args):
        ...
        self.session_arguments = {
            "verify": self.session.rejectUnauthorized,
            "timeout": 30,
            **session_args
        }
```

t1m0thyj (changes requested):
@savannah-portilla1 Thanks for your contribution!

Left a comment on the PR. But first please fix the failing DCO check - see [here](https://github.com/zowe/zowe-client-python-sdk/pull/178/checks?check_run_id=13864998256) for how to do so, and it will be easiest to fix if you do it before making more commits.

Issue #63, opened 2022-06-07
[zos-files] Add method to copy data set or member


t1m0thyj:
Add a method to the Python zos-files SDK to call the z/OSMF REST API `PUT /zosmf/restfiles/ds/<to-data-set-name>` with `copy` request

* z/OSMF REST API documentation: https://www.ibm.com/docs/en/zos/2.5.0?topic=interface-zos-data-set-member-utilities
* Python zos-files SDK in the "main" branch: https://github.com/zowe/zowe-client-python-sdk/tree/main/src/zos_files
* Node.js zos-files SDK for comparison: https://github.com/zowe/zowe-cli/blob/master/packages/zosfiles/src/methods/copy/Copy.ts

samadpls:
Is this issue still active? May I proceed with addressing it?

t1m0thyj:
@samadpls I think you already addressed in #184 but GitHub didn't auto-close the issue when the PR was merged. Thanks for following up 🙂

PR #184, opened 2023-06-08
fixed issue #63 Added methods to copy (dataset or member) & USS files

samadpls:
**What It Does**

add method `copy_dataset_or_member`
Resolve issue #63  

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+            to_dataset_name: str
+                Name of the dataset to copy to
+            to_member_name: str
+                Name of the member to copy to  
+            type: str
+                Type of the file to copy from
+            enq: str
+                Enqueue type for the file to copy from
+            replace: bool
+                If true, the function uses the REPLACE=YES on ARCHDEF request, otherwise it uses the REPLACE=NO.
```
t1m0thyj, line 159: Not everyone will be familiar with what an ARCHDEF request is. For the replace option I think we'll want to say something like the following:
`If true, members in the target data set are replaced.`

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+        data={
+            "request":"copy",
+           "from-file":{
+               "filename":from_filname.strip(),
+                "type":type.strip()
+           },
+           "replace":replace
+        }
+        
+        path_to_member = f"{to_dataset_name.strip()}({to_member_name.strip()})" if to_member_name else to_dataset_name.strip()
```
t1m0thyj, line 175: I don't think it's necessary to `strip` all of the input strings. We can either assume that the user has provided strings without extra spaces, or that if there are any the REST API will handle them.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+        
+        path_to_member = f"{to_dataset_name.strip()}({to_member_name.strip()})" if to_member_name else to_dataset_name.strip()    
+        custom_args = self._create_custom_request_arguments()
+        custom_args['json'] = data
+        custom_args["url"] = "{}ds/{}".format(self.request_endpoint, path_to_member)
+        response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[200])
+        return response_json
+    
+    def copy_dataset_or_member(self,from_dataset_name,to_dataset_name,from_member_name="",volser="",alias="",
+                               to_member_name="",enq="",replace=False):
```
t1m0thyj, line 183: For the default value of `from_member_name`, `to_member_name`, `volser`, `alias`, and `enq`, can we use `None` instead of an empty string?

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+            volser: str
+                Volume serial number of the dataset to copy from
+            alias: str  
+                Alias of the dataset to copy from
+            to_member_name: str
+                Name of the member to copy to
+            enq: str
+                Enqueue type for the dataset to copy from
+            replace: bool
+                If true, the function uses the REPLACE=YES on ARCHDEF request, otherwise it uses the REPLACE=NO.
```
t1m0thyj, line 203: Same comment as above about the docs for the `replace` option

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+        path_to_member = f"{to_dataset_name.strip()}({to_member_name.strip()})" if to_member_name else to_dataset_name.strip()
+        if enq:
+            if enq in ("SHR","SHRW","EXCLU"):
+                data["enq"] = enq.strip()
+            else:
+                raise ValueError("Invalid value for enq.")
+        if volser:
+             data["from-dataset"]["volser"]=volser.strip()
+        if alias:
+            data["from-dataset"]["alias"]=alias.strip()
```
t1m0thyj, line 229: Not necessary to strip all of the input strings (same comment as above).

t1m0thyj (changes requested):
Thanks @samadpls 🙂 Left a few comments about the function parameters

samadpls:
@t1m0thyj , thank you for your comments. I have updated the PR.

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -139,7 +139,101 @@ def list_dsn_members(self, dataset_name, member_pattern=None,
...
+        
+        path_to_member = f"{to_dataset_name}({to_member_name})" if to_member_name else to_dataset_name
+        custom_args = self._create_custom_request_arguments()
+        custom_args['json'] = data
+        custom_args["url"] = "{}ds/{}".format(self.request_endpoint, path_to_member)
+        response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[200])
+        return response_json
+    
+    def copy_dataset_or_member(self,from_dataset_name,to_dataset_name,from_member_name=None,volser=None,alias=None,
+                               to_member_name=None,enq=None,replace=False):
```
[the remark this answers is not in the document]
t1m0thyj: Discussed with @zFernand0, and we do want keep all the parameters as they were defined before.

Could you please restore the full method signature?
`def copy_dataset_or_member(self, from_dataset_name, to_dataset_name, from_member_name=None, volser=None, alias=None, to_member_name=None, enq=None, replace=False):`

In the tests, we can continue using a dictionary to define the keyword args 👍

t1m0thyj (approved):

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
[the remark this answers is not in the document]
zFernand0: Sorry for not commenting sooner. And thanks for summarizing the discussion

zFernand0 (approved):

PR #189, opened 2023-06-14
Update workflows and document installing from source

t1m0thyj:
**What It Does**
* Update branch names in workflows and readme links
* Disable "dev release" workflow until we're ready to publish to PyPI
* Configure `setup.py` files to allow installing from source

**How to Test**
Test the command in the readme to install from GitHub, but using this branch instead of main:
`pip install zowe@git+https://github.com/zowe/zowe-client-python-sdk@update-workflows#subdirectory=src`

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**
We can probably squash merge this PR since the commit history is kind of ugly 😁


src/zosmf/setup.py
```diff
@@ -1,8 +1,21 @@
...
+src_dir = os.path.realpath(os.path.dirname(os.path.dirname(__file__)))
+
+def resolve_sdk_dep(sdk_name, version_spec):
+    if os.path.exists(os.path.join(src_dir, sdk_name, "zowe")):
+        # Handle building from a Git checkout
+        # Based on https://github.com/lab-cosmo/equistore/blob/master/python/equistore-torch/setup.py#L212
+        sdk_dir = os.path.realpath(os.path.join(src_dir, sdk_name))
+        return f"zowe.{sdk_name}_for_zowe_sdk@file://{sdk_dir}?{uuid.uuid4()}"
+    else:
+        return f"zowe.{sdk_name}_for_zowe_sdk{version_spec}"
```
traeok, line 17: Can this be separated into a common script/file that we can reference in all the `setup.py`'s?

traeok (approved):

zFernand0 (approved):
code duplication can always be address at a later time 😋

traeok (approved):
This looks good, thanks @t1m0thyj! just left a comment about the README.

I'm curious, will the use of `sys.path.insert(0` introduce any bugs? I was reading into the Python doc and it seems like the path is prepended with a potentially unsafe path:

> `python -c code` and python (REPL) command lines: prepend an empty string, which means the current working directory.
> `python script.py` command line: prepend the script’s directory. If it’s a symbolic link, resolve symbolic links.
> `python -m module` command line: prepend the current working directory.
> By default, as initialized upon program startup, a potentially unsafe path is prepended to [sys.path](https://docs.python.org/3/library/sys.html#sys.path) (before the entries inserted as a result of [PYTHONPATH](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH))

If we are intentionally modifying this path (which it seems like we are for properly routing packages), then this comment can be discarded. I just want to make sure its not causing any issues with the way the Python scripts will run.

t1m0thyj:
> This looks good, thanks @t1m0thyj! just left a comment about the README.
> 
> I'm curious, will the use of `sys.path.insert(0` introduce any bugs when running the scripts? I was reading into the Python doc and it seems like the path is prepended with a potentially unsafe path. Since the line was only added to the setup Python files, this can probably be ignored. I found this under the `sys.path` docs for reference:
> 
> > `python -c code` and python (REPL) command lines: prepend an empty string, which means the current working directory.
> > `python script.py` command line: prepend the script’s directory. If it’s a symbolic link, resolve symbolic links.
> > `python -m module` command line: prepend the current working directory.
> > By default, as initialized upon program startup, a potentially unsafe path is prepended to [sys.path](https://docs.python.org/3/library/sys.html#sys.path) (before the entries inserted as a result of [PYTHONPATH](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH))
> 
> If we are intentionally modifying this path (which it seems like we are for properly routing packages), then this comment can be discarded. I just want to make sure its not causing any issues with the way the Python scripts will run.

I think for setup.py we don't need to worry about negative side effects since its just a packaging script. The reason for using `sys.path.insert` was that without the parent directory first in the list, trying to import `from setup` in the lower level setup.py files causes a circular import because it will try to import itself from the current directory.

zFernand0:
LGTM! 😋

PR #191, opened 2023-06-23
Implementws secure value loading method for multiple Windows credentials

### LINKED ISSUE #134: Load secure values from multiple credential entries on Windows
On Windows the maximum credential length is 2560 bytes. If the JSON object in `Zowe/secure_config_props` exceeds this length, Zowe CLI works around the limitation by splitting the credential string across multiple entries. The final one is terminated by a NUL byte (`\0`). For example:
**Zowe/secure_config_props-1**
```
{"C:/Users/User/.zowe/zowe.con
```
**Zowe/secure_config_props-2**
```
fig.json":{...}}␀
```

samadpls:
**What It Does**

resolved #134 
**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/zosmf_profile.py
+++ b/src/core/zowe/core_for_zowe_sdk/zosmf_profile.py
@@ -119,11 +119,32 @@ class ZosmfProfile:
         try:
             zosmf_user = self.__get_secure_value("user")
             zosmf_password = self.__get_secure_value("password")
+            if sys.platform == "win32":
+                if len(zosmf_user) > constants["WIN32_CRED_MAX_STRING_LENGTH"]:
+                    # Split user value across multiple fields
+                    self.__split_and_store_secure_value("user", zosmf_user)
+                if len(zosmf_password) > constants["WIN32_CRED_MAX_STRING_LENGTH"]:
+                    # Split password value across multiple fields
+                    self.__split_and_store_secure_value("password", zosmf_password)    
         except Exception as e:
             raise SecureProfileLoadFailed(self.profile_name, e)
         else:
             return (zosmf_user, zosmf_password)
+        
+    def __split_and_store_secure_value(self, field_name, value):
+        """Split and store a secure value across multiple fields."""
+        service_name = constants["ZoweCredentialKey"]
+        account_name = f"zosmf_{self.profile_name}_{field_name}"
+        
+        # Delete any previously used fields storing this value
+        keyring.delete_password(service_name, account_name)
+        sliced_values = [value[i:i+constants["WIN32_CRED_MAX_STRING_LENGTH"]] for i in range(0, len(value), constants["WIN32_CRED_MAX_STRING_LENGTH"])]
+    
+        for index, temp_value in enumerate(sliced_values, start=1):
+            keyring.set_password(service_name, f"{account_name}-{index}", temp_value)
 
+        # Append null byte to mark the termination of the last field
+        keyring.set_password(service_name, f"{account_name}-{len(sliced_values)+1}", "\0")
 
 if HAS_KEYRING and sys.platform.startswith("linux"):
     from contextlib import closing
--- a/src/core/zowe/core_for_zowe_sdk/constants.py
+++ b/src/core/zowe/core_for_zowe_sdk/constants.py
@@ -16,4 +16,5 @@ constants = {
     "ZoweCredentialKey": "Zowe-Plugin",
     "ZoweServiceName": "Zowe",
     "ZoweAccountName": "secure_config_props",
+    "WIN32_CRED_MAX_STRING_LENGTH" : 2560 
 }


src/core/zowe/core_for_zowe_sdk/zosmf_profile.py
t1m0thyj, line 1: This file (zosmf_profile.py) contains code for loading the old Zowe v1 format of profiles which were stored as YAML files in the `~/.zowe/profiles` directory. Zowe v1 stored a separate credential in the vault for each profile property (user, password, etc) so the length never exceeded the 2560 character limit.

The code for loading secure values for Zowe v2 team config files is located in config_file.py. Now all of a user's secure values are stored in a single credential in the vault. Storing in a single credential is beneficial for platforms like MacOS where users may have to enter a password for each credential that is loaded, but the length may easily exceed 2560 characters.

Could you please move this logic for splitting the secure values into config_file.py? Sorry for not providing more context about this when we discussed the issue 😢

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +341,76 @@ def load_secure_props(self) -> None:
...
+                        field_name = f"{constants['ZoweAccountName']}-{index}"
+                        keyring.set_password(service_name, field_name, chunk)
+                else:
+                    # Credential length is within the maximum limit, set it as a single keyring entry
+                    keyring.set_password(service_name, constants["ZoweAccountName"], credential)
+            else:
+                credential = self.secure_props.get(self.filepath)
+                keyring.set_password(
+                    service_name, constants["ZoweAccountName"], 
+                    credential)
```
t1m0thyj, line 408: The `secure_config_props` object that we store in the vault contains credentials for multiple file paths, for example:
```json
{
  "/Users/Timothy/.zowe/zowe.config.json": {
    "profiles.base.properties.user": "abc",
    "profiles.base.properties.password": "def"
  },
  "/Users/Timothy/Projects/sample-zowe-project/zowe.config.json": {
    "profiles.base.properties.user": "ghi",
    "profiles.base.properties.password": "jkl"
  }
}
```

If we store the secure properties for only a single file, then we'll end up overwriting and losing secure properties for any other files. So I think we want to do something like this in `set_secure_props`:
* Load credentials from the vault again in case they changed from the version we have cached in memory
* Update secure properties for the current config file's path
* Save the entire credential object back to the vault

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,98 @@ def load_secure_props(self) -> None:
...
+            if credential:
+                # Get the username and password from the credential dictionary
+                username = credential.get("profiles.base.properties.user")
+                password = credential.get("profiles.base.properties.password")
+                
+                # Combine the username and password as "username:password"
+                username_password = f"{username}:{password}"
+                
+                # Encode the combined string as base64
+                encoded_credential = base64.b64encode(username_password.encode()).decode()
```
t1m0thyj, line 405: Please remove this logic - we don't want to be requiring a user and password on the base profile as they are optional.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,98 @@ def load_secure_props(self) -> None:
...
+                if sys.platform == "win32":
+                    service_name += "/" + constants["ZoweAccountName"]
+                    
+                    # Load existing credentials
+                    existing_credential = keyring.get_password(service_name, constants["ZoweAccountName"])
```
t1m0thyj, line 410: We should call the `_retrieve_password` helper function here in case the credentials are split across multiple vault entries on Windows.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,98 @@ def load_secure_props(self) -> None:
...
+                    if existing_credential:
+                        # Decode the existing credential and update secure_props
+                        existing_secure_props = commentjson.loads(existing_credential)
+                        existing_secure_props.update(self.secure_props.get(self.filepath, {}))
+                        self.secure_props[self.filepath] = existing_secure_props
```
t1m0thyj, line 415: The `_retrieve_password` function can handle JSON parsing and Base64-decoding for us.

I think this code could be refactored into something like this:
```
existing_secure_props[self.filepath] = self.secure_props
# stringify the existing_secure_props object
# base64 encode the string to get the "encoded_credential"
```

t1m0thyj (changes requested):

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,94 @@ def load_secure_props(self) -> None:
...
+        Returns
+        -------
+        None
+        """
+        if not HAS_KEYRING:
+            return
+
+        try:
+            service_name = constants["ZoweServiceName"]
+            credential = self.secure_props.get(self.filepath, {})
```
t1m0thyj, line 394: We want to store the entire credentials object which has file paths as keys:
`credential = {self.filepath: self.secure_props}`

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,94 @@ def load_secure_props(self) -> None:
...
+                    service_name += "/" + constants["ZoweAccountName"]
+                    
+                    # Load existing credentials
+                    existing_credential = self._retrieve_credential(service_name)
+                    if existing_credential:
+                        # Decode the existing credential and update secure_props
+                        existing_credential_bytes = base64.b64decode(existing_credential.encode())
+                        existing_secure_props = commentjson.loads(existing_credential_bytes)
+                        existing_secure_props[self.filepath] = self.secure_props[self.filepath]
+                        encoded_credential = base64.b64encode(commentjson.dumps(existing_secure_props).encode()).decode()
```
t1m0thyj, line 410: This logic should be moved up outside of the `if sys.platform == "win32"` condition, because on all platforms we want to merge credentials for this file into the credentials object that may already exist in the keyring.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +345,94 @@ def load_secure_props(self) -> None:
...
+
+                    
+                        if len(encoded_credential) > constants["WIN32_CRED_MAX_STRING_LENGTH"]:
+                            # Split the encoded credential string into chunks of maximum length
+                            keyring.delete_password(service_name, constants["ZoweAccountName"])
```
t1m0thyj, line 415: We need to check if the password already exists with `keyring.get_password` first - otherwise the call to `keyring.delete_password` will fail.

t1m0thyj (changes requested):
Left a few comments on the `set_secure_props` method.

Before merging, we should do manual tests on Windows to make sure that `set_secure_props` stores very long credentials as expected for a zowe.config.json file. We can discuss on Slack if you have any questions about how to test 🙂

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -309,27 +310,25 @@ def load_secure_props(self) -> None:
 
         try:
             service_name = constants["ZoweServiceName"]
+            is_win32 = sys.platform == "win32"
 
-            if sys.platform == "win32":
+            if is_win32:
                 service_name += "/" + constants["ZoweAccountName"]
```
t1m0thyj, line 316: This code could be moved inside the `_retrieve_credential` method along with other Windows specific logic 🙂

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +339,135 @@ def load_secure_props(self) -> None:
...
+                temp_value = keyring.get_password(f"{service_name}-{index}", f"{constants['ZoweAccountName']}-{index}")
+                
+        if encoded_credential is not None and encoded_credential.endswith("\0"):
+            encoded_credential = encoded_credential[:-1]
+
+        try:
+            return encoded_credential.encode('utf-16le').decode()
+        except (UnicodeDecodeError, AttributeError):
+            # The credential is not encoded in UTF-16
+            return encoded_credential
```
t1m0thyj, line 379: I believe we only need this logic on Windows, as other platforms are less weird and don't encode the credentials as UTF-16.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +339,135 @@ def load_secure_props(self) -> None:
...
+                else:
+                    print("here")
+                    # Encode the credential
+                    encoded_credential = base64.b64encode(commentjson.dumps(credential).encode()).decode() 
+                print(encoded_credential)  
+                # Check if the encoded credential exceeds the maximum length for win32
+                if is_win32 and len(encoded_credential) > constants["WIN32_CRED_MAX_STRING_LENGTH"]:
+                    # Split the encoded credential string into chunks of maximum length
+                    chunk_size = constants["WIN32_CRED_MAX_STRING_LENGTH"]
+                    chunks = [encoded_credential[i: i + chunk_size] for i in range(0, len(encoded_credential), chunk_size)]
```
t1m0thyj, line 454: If the credential exceeds the maximum length on Windows, then we need to add a `\0` byte at the end to indicate end of data before we split it into chunks.

This is separate from the second `\0` that is already present inside the for loop to ensure the string is an even number of characters.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +339,135 @@ def load_secure_props(self) -> None:
...
+            The account name for the credential deletion
+        Returns
+        -------
+        None
+        """
+        
+        try:
+            keyring.delete_password(service_name, account_name)
+        except keyring.errors.PasswordDeleteError:
+            # keyring.delete_password is not available (e.g., macOS)
```
t1m0thyj, line 399: Curious about this comment, since I'm on MacOS and `keyring.delete_password` works for me 🤔 

No need to change anything here though, the error handling looks good 👍

t1m0thyj (commented):
The new CredentialManager implementation is looking good 👍  I like the usage of `warnings.filterwarnings` - clever way to prevent the keyring package from spamming the console 🙂 

It looks like most of the methods for credential management are duplicated in `config_file.py` and `credential_manager.py`. I assume this is temporary so that tests will pass, and the next step would be to consolidate into a single implementation in the CredentialManager class?

src/core/zowe/core_for_zowe_sdk/config_file.py
samadpls: Yes, you are right, `keyring.delete` works fine on Mac Os. I'll update the comment. Here, what I meant was this next logic only applicable for Windows.

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -0,0 +1,195 @@
...
+        str
+            The retrieved  encoded credential
+        """
+        is_win32 = sys.platform == "win32"
+        if is_win32:
+            service_name += "/" + constants["ZoweAccountName"]
+        encoded_credential = keyring.get_password(service_name, constants["ZoweAccountName"])
+        if encoded_credential is None and is_win32:
+            # Filter or suppress specific warning messages
+            warnings.filterwarnings("ignore", message="^Retrieved an UTF-8 encoded credential")
```
t1m0thyj, line 85: The first call to `keyring.get_password` may display an encoding warning on Windows.
I think the `filterwarnings` call should be moved up into the previous if block so that it runs first.

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -0,0 +1,195 @@
...
+                CredentialManager.delete_credential(service_name , constants["ZoweAccountName"])
+            else:
+                # Encode the credential
+                encoded_credential = base64.b64encode(commentjson.dumps(credential).encode()).decode() 
+            # Check if the encoded credential exceeds the maximum length for win32
+            if is_win32 and len(encoded_credential) > constants["WIN32_CRED_MAX_STRING_LENGTH"]:
+                # Split the encoded credential string into chunks of maximum length
+                chunk_size = constants["WIN32_CRED_MAX_STRING_LENGTH"]
+                chunks = [encoded_credential[i: i + chunk_size] for i in range(0, len(encoded_credential), chunk_size)]
+                chunks[-1]+= '\0'
```
t1m0thyj, line 184: If the password is 5120 (2560*2) bytes, this will make the last chunk 2561 bytes which causes an error.
This can be fixed by adding the trailing `\0` byte before splitting the password in chunks instead of after.

t1m0thyj (changes requested):

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -0,0 +1,195 @@
...
+            if existing_credential:
+                    
+                # Decode the existing credential and update secure_props
+                existing_credential_bytes = base64.b64decode(existing_credential).decode()
+                existing_secure_props = commentjson.loads(existing_credential_bytes)
+                existing_secure_props.update(credential)
+                # Encode the credential
+                encoded_credential = base64.b64encode(commentjson.dumps(existing_secure_props).encode()).decode()
+                # Delete the existing credential
+                CredentialManager.delete_credential(service_name , constants["ZoweAccountName"])
```
t1m0thyj, line 175: Previously we were storing secure properties for only a single config file in memory, and had to rebuild the entire object. Now that CredentialManager stores the entire secure properties object, we don't need to load the existing credential anymore.

Therefore I think these lines can be removed 🙂 Except on Windows, we should delete the credential if it already exists, to clean up old properties that may have a different name (like `Zowe/secure_config_props-1`). This logic could be added to the `if is_win32` block above.

src/core/zowe/core_for_zowe_sdk/config_file.py
t1m0thyj, line 1: Please remove unused imports (`keyring`, `base64`, etc) from the top of this file.

t1m0thyj (approved):
LGTM and thanks for adding tests 👍 

Just have one comment about unused imports 😋

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -0,0 +1,183 @@
...
+        ----------
+        service_name: str
+            The service name for the credential retrieval
+        Returns
+        -------
+        str
+            The retrieved  encoded credential
+        """
+        # Configure the logger to ignore warning messages
+        logging.getLogger().setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
t1m0thyj: I think `warnings.filterwarnings` would be ideal here if we can get it to work, since we can be specific about the warnings we want to filter out 👍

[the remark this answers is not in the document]
samadpls: I'm not sure why, but the `filterwarning` were not able to hide the warning in the terminal 😥

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -0,0 +1,183 @@
...
+        None
+        """
+        
+        try:
+            keyring.delete_password(service_name, account_name)  
+        except keyring.errors.PasswordDeleteError:
+            # Handling multiple credentials stored when the operating system is Windows
+            if sys.platform == "win32":
+                # Delete the secure value with an index
+                index = 1
```
zFernand0, line 136: We may want to do the multi-credential handling outside of the `except` to make sure that all `service/account` are deleted 😋

samadpls: I believe if the service and account name do not exist, which is the case when it's stored in Credential Manager in a multi-credential way, and when we call it, it gives a PasswordDeleteError. I apologize if I didn't understand your comment correctly.😥

zFernand0: You are correct.
I was mostly wondering in the event that there is a combination of both, multi-credentials and regular ones in Windows.
If that's the case, we won't be able to remove the `-1, -2, ...` the same way that the Nodejs SDK would do it.

It's kind of a minor detail though 😋

src/core/zowe/core_for_zowe_sdk/credential_manager.py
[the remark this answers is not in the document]
t1m0thyj: Created #207

zFernand0 (approved):
LGTM! 😋

PR #192, opened 2023-06-26
Validate that zowe.config.json file matches schema

### LINKED ISSUE #74: [core] Validate that zowe.config.json file matches schema

aadityasinha-dotcom:
**What It Does**
Resolves #74 

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -118,7 +118,7 @@ class ConfigFile:
     def init_from_file(
         self, 
         config_type: str, 
-        opt_in: Optional[bool] = True
+        opt_out: Optional[bool] = True
     ) -> None:
         """
         Initializes the class variable after
@@ -135,7 +135,7 @@ class ConfigFile:
         self.defaults = profile_jsonc.get("defaults", {})
 
         if self.schema_property:
-            self.validate_schema(config_type, opt_in)
+            self.validate_schema(config_type, opt_out)
         # loading secure props is done in load_profile_properties
         # since we want to try loading secure properties only when
         # we know that the profile has saved properties
@@ -144,7 +144,7 @@ class ConfigFile:
     def validate_schema(
         self,
         config_type: str,
-        opt_in: Optional[bool] = True,
+        opt_out: Optional[bool] = True,
     ) -> None:
         """
         Get the $schema_property from the config and load the schema
@@ -166,7 +166,7 @@ class ConfigFile:
             )
                 
         # validate the $schema property 
-        if path_schema_json and opt_in:
+        if path_schema_json and opt_out:
             validate_config_json(path_config_json, path_schema_json)
         
 
@@ -175,7 +175,7 @@ class ConfigFile:
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
         config_type: Optional[str] = None,
-        opt_in: Optional[bool] = True,
+        opt_out: Optional[bool] = True,
     ) -> Profile:
         """
         Load given profile including secure properties and excluding values from base profile
@@ -185,7 +185,7 @@ class ConfigFile:
             Returns a namedtuple called Profile
         """
         if self.profiles is None:
-            self.init_from_file(config_type, opt_in)
+            self.init_from_file(config_type, opt_out)
 
         if profile_name is None and profile_type is None:
             raise ProfileNotFound(
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -99,7 +99,7 @@ class ConfigFile:
     @property
     def location(self) -> Optional[str]:
         return self._location
-    
+
     @property
     def schema_path(self) -> Optional[str]:
         return self.schema_property
@@ -112,7 +112,7 @@ class ConfigFile:
             raise FileNotFoundError(f"given path {dirname} is not valid")
 
     def init_from_file(
-        self, 
+        self,
         validate_schema: Optional[bool] = True,
     ) -> None:
         """
@@ -155,22 +155,22 @@ class ConfigFile:
             warnings.warn(
                 f"$schema property could not found"
             )
-                
-        # validate the $schema property 
+
+        # validate the $schema property
         if path_schema_json:
             validate_config_json(self.jsonc, path_schema_json, cwd = self.location)
-        
+
     def schema_list(
         self,
     ) -> list:
         """
         Loads the schema properties
         in a sorted order according to the priority
-        
+
         Returns
         -------
         Dictionary
-        
+
             Returns the profile properties from schema (prop: value)
         """
 
@@ -181,12 +181,12 @@ class ConfigFile:
         if schema.startswith("https://") or schema.startswith("http://"):
             schema_json = requests.get(schema).json()
 
+        elif os.path.isfile(schema) or schema.startswith("file://"):
+            with open(schema.replace("file://", "")) as f:
+                schema_json = json.load(f)
+
         elif not os.path.isabs(schema):
             schema = os.path.join(self.location, schema)
-            with open(schema) as f:
-                schema_json = json.load(f)
-        
-        elif os.path.isfile(schema):
             with open(schema) as f:
                 schema_json = json.load(f)
         else:
@@ -194,10 +194,10 @@ class ConfigFile:
 
         profile_props:dict = {}
         schema_json = dict(schema_json)
-        
+
         for props in schema_json['properties']['profiles']['patternProperties']["^\\S*$"]["allOf"]:
             props = props["then"]
-            
+
             while "properties" in props:
                 props = props.pop("properties")
                 profile_props = props
@@ -303,7 +303,7 @@ class ConfigFile:
             profile_name=profile_type,
             error_msg=f"No profile with matching profile_type '{profile_type}' found",
         )
-        
+
     def find_profile(self, path: str, profiles: dict):
         """
         Find a profile at a specified location from within a set of nested profiles
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -113,7 +113,6 @@ class ConfigFile:
 
     def init_from_file(
         self, 
-        config_type: str, 
         validate_schema: Optional[bool] = True,
     ) -> None:
         """
@@ -209,7 +208,6 @@ class ConfigFile:
         self,
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
-        config_type: Optional[str] = None,
         validate_schema: Optional[bool] = True,
     ) -> Profile:
         """
@@ -220,7 +218,7 @@ class ConfigFile:
             Returns a namedtuple called Profile
         """
         if self.profiles is None:
-            self.init_from_file(config_type, validate_schema)
+            self.init_from_file(validate_schema)
 
         if profile_name is None and profile_type is None:
             raise ProfileNotFound(
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -121,7 +121,7 @@ class ConfigFile:
     def init_from_file(
         self, 
         config_type: str, 
-        validate_schema: Optional[bool] = True
+        validate_schema: Optional[bool] = False
     ) -> None:
         """
         Initializes the class variable after
@@ -148,7 +148,7 @@ class ConfigFile:
     def validate_schema(
         self,
         config_type: str,
-        validate_schema: Optional[bool] = True,
+        validate_schema: Optional[bool] = False,
     ) -> None:
         """
         Get the $schema_property from the config and load the schema
@@ -222,7 +222,7 @@ class ConfigFile:
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
         config_type: Optional[str] = None,
-        validate_schema: Optional[bool] = True,
+        validate_schema: Optional[bool] = False,
     ) -> Profile:
         """
         Load given profile including secure properties and excluding values from base profile
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -115,7 +115,6 @@ class ConfigFile:
         self, 
         config_type: str, 
         validate_schema: Optional[bool] = True,
-        verify: Optional[bool] = False,
     ) -> None:
         """
         Initializes the class variable after
@@ -133,16 +132,14 @@ class ConfigFile:
         self.jsonc = profile_jsonc
 
         if self.schema_property and validate_schema:
-            self.validate_schema(config_type, validate_schema)
+            self.validate_schema()
         # loading secure props is done in load_profile_properties
         # since we want to try loading secure properties only when
         # we know that the profile has saved properties
         # self.load_secure_props()
 
     def validate_schema(
-        self,
-        config_type: str,
-        verify: Optional[bool] = False,
+        self
     ) -> None:
         """
         Get the $schema_property from the config and load the schema
@@ -152,10 +149,7 @@ class ConfigFile:
         file_path to the $schema property
         """
 
-        path_schema_json, path_config_json = None, ""
-
-        if self._location:
-            path_config_json = f"./{self.filename}"
+        path_schema_json = None
 
         path_schema_json = self.schema_path
         if path_schema_json is None:    # check if the $schema property is not defined
@@ -165,7 +159,7 @@ class ConfigFile:
                 
         # validate the $schema property 
         if path_schema_json:
-            validate_config_json(self.jsonc, path_schema_json, cwd = self.location, verify=verify)
+            validate_config_json(self.jsonc, path_schema_json, cwd = self.location)
         
     def schema_list(
         self,
@@ -217,7 +211,6 @@ class ConfigFile:
         profile_type: Optional[str] = None,
         config_type: Optional[str] = None,
         validate_schema: Optional[bool] = True,
-        verify: Optional[bool] = False,
     ) -> Profile:
         """
         Load given profile including secure properties and excluding values from base profile
@@ -227,7 +220,7 @@ class ConfigFile:
             Returns a namedtuple called Profile
         """
         if self.profiles is None:
-            self.init_from_file(config_type, validate_schema, verify)
+            self.init_from_file(config_type, validate_schema)
 
         if profile_name is None and profile_type is None:
             raise ProfileNotFound(
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -102,6 +102,10 @@ class ConfigFile:
     @property
     def location(self) -> Optional[str]:
         return self._location
+    
+    @property
+    def schema_path(self) -> Optional[str]:
+        return self.schema_property
 
     @location.setter
     def location(self, dirname: str) -> None:
@@ -296,19 +300,6 @@ class ConfigFile:
             #     self._missing_secure_props.extend(secure_fields)
 
         return props
-
-    def load_schema(self) -> str:
-        """
-        Find the schema file path from the $schema_property in the config.json
-        Returns
-        -------
-        Str
-
-        file_path to the schema property
-        """
-
-        file_path = self.schema_property
-        return file_path
 
     def load_secure_props(self) -> None:
         """
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -69,6 +69,7 @@ class ConfigFile:
     4. Contents of the file.
     4.1 Profiles
     4.2 Defaults
+    4.3 Schema Property
     5. Secure Properties associated with the file.
     """
 
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -111,10 +111,6 @@ class ConfigFile:
     def schema_path(self) -> Optional[str]:
         return self.schema_property
 
-    @property
-    def schema_path(self) -> Optional[str]:
-        self.schema_property
-
     @location.setter
     def location(self, dirname: str) -> None:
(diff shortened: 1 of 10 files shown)


aadityasinha-dotcom:
@t1m0thyj @zFernand0 I have created the draft PR as this is currently incomplete. Need some reviews on it, and how we will handle/find the errors while doing validation :blush:

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -120,6 +122,7 @@ def init_from_file(self) -> None:
             profile_jsonc = commentjson.load(fileobj)
 
         self.profiles = profile_jsonc.get("profiles", {})
+        self.schema_property = profile_jsonc.get("$schema", None)
```
aadityasinha-dotcom, line 125: I have added the `schema_property` in `ConfigFile` class, so that it will automatically search for the `$schema` property

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -169,12 +171,57 @@ def get_profile(
...
+                f"Validating a format {path_config_json} failed for {path_schema_json}"
+            )
+        except jsonschema.UndefinedTypeCheck as exc:
+            warnings.warn(
+                f"A type checker was asked to check a type it did not have registered, {exc}"
+            )
+        except jsonschema.UnknownType:
+            warnings.warn(
+                f"Unknown type is found in {path_schema_json}"
+            )
```
aadityasinha-dotcom, line 215: Here are the errors that can be handled

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -224,6 +272,11 @@ def load(
                 )  # Define profile name that will be merged from other layers
             profile_props = {**profile_loaded.data, **profile_props}
 
+            # Loading $schema property from project config
+            if config_type in ("Project Config"):
+                path_config_json = cfg._location + "/zowe.config.json"
+                schema_path = self.load_schema(cfg, path_config_json, opt_in)
```
aadityasinha-dotcom, line 278: I don't know if it is require to store the output in some variable or just call it and do not return anything from the function

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -169,12 +171,57 @@ def get_profile(
...
+            )
+        except jsonschema.UndefinedTypeCheck as exc:
+            raise Exception(
+                f"A type checker was asked to check a type it did not have registered, {exc}"
+            )
+        except jsonschema.UnknownType:
+            raise Exception(
+                f"Unknown type is found in {path_schema_json}"
+            )
+        finally:
```
aadityasinha-dotcom, line 216: (Updated) Raised errors for validation fails

src/core/zowe/core_for_zowe_sdk/profile_manager.py
t1m0thyj: I don't think we need to store the output in a variable in this file. It should be sufficient to have `schema_path` defined as a property of the ConfigFile class.

Can we validate the schema for all config layers rather than only for "Project Config"?

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -169,12 +171,57 @@ def get_profile(
...
+        try:
+            path_schema_json = cfg.load_schema()
+            if path_schema_json is None:    # check if the $schema property is not defined
+                warnings.warn(
+                    f"$schema property could not found"
+                )
+                
+            # validate the $schema property 
+            if path_schema_json and opt_in:
+                validate_config_json(path_config_json, path_schema_json)
```
t1m0thyj, line 199: Can we enhance the `validate_config_json` method to download the schema from the internet if `path_schema_json` points to an Internet URI rather than local file path?

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -224,6 +274,14 @@ def load(
                 )  # Define profile name that will be merged from other layers
             profile_props = {**profile_loaded.data, **profile_props}
 
+            # Validating $schema property for all the layers
+            if cfg._location and config_type in ("Project Config", "Global Config"):
+                path_config_json = cfg._location + "/zowe.config.json"
+                self.validate_schema(cfg, path_config_json, opt_in)
+            elif cfg._location and config_type in ("Project User Config", "Global User Config"):
+                path_config_json = cfg._location + "/zowe.config.user.json"
+                self.validate_schema(cfg, path_config_json, opt_in)
```
t1m0thyj, line 283: I don't think validating the schema needs to be specific to the ProfileManager class. We could probably move the `validate_schema` method into the ConfigFile class, since its an action that applies to any individual config file. @zFernand0 do you agree? 😋

src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -31,10 +33,22 @@ def validate_config_json(path_config_json: str, path_schema_json: str):
         Provides details if config.json doesn't match schema.json, otherwise it returns None.
     """
 
+    # checks if the path_schema_json point to an internet URI and download the schema using the URI
+    if path_schema_json[:8] in ("https://", "http://"):
```
t1m0thyj, line 37: Grabbing the first 8 characters here won't always work. `https://` is 8 characters long but `http://` is only 7.

src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -31,10 +33,22 @@ def validate_config_json(path_config_json: str, path_schema_json: str):
...
+    elif os.path.isfile(path_schema_json):
+        with open(path_schema_json) as file:
+            schema_json = commentjson.load(file)
+        
+    # if there is no path_schema_json it will return None
+    else: 
+        return None
+    
     with open(path_config_json) as file:
         config_json = commentjson.load(file)
```
t1m0thyj, line 52: Currently we are loading the config file with `commentjson` twice. We could increase performance by passing the loaded JSON object directly into `validate_config_json`. Perhaps something like this:
```python
def validate_config_json(path_config_json: Union[str, dict], path_schema_json: str):
    if isinstanceof(path_config_json, str):
        # load config from file name
    else:
        # use config object passed in
```

src/core/zowe/core_for_zowe_sdk/validators.py
t1m0thyj: When you write unit tests for this method, please include one that validates JSON using an online schema like the following: https://json-schema.org/draft/2020-12/schema

src/core/zowe/core_for_zowe_sdk/profile_manager.py
zFernand0: it seems that this was already addressed 😋

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -120,17 +131,53 @@ def init_from_file(self) -> None:
...
+        -------
+        file_path to the $schema property
+        """
+
+        path_schema_json, path_config_json = None, ""
+
+        if self._location and config_type in ("Project Config", "Global Config"):
+            path_config_json = "./zowe.config.json"
+        elif self._location and config_type in ("Project User Config", "Global User Config"):
+            path_config_json = "./zowe.config.user.json"
```
t1m0thyj, line 162: I think this could be simplified to `path_config_json = f"./{self.filename}"`, since the `filename` property returns either "zowe.config.json" or "zowe.config.user.json" depending on the config type.

aadityasinha-dotcom: Thanks for the review, anything else left for the changes :relaxed:

tests/unit/test_zowe_core.py
```diff
@@ -380,6 +384,27 @@ def test_profile_not_found_warning(self, get_pass_func):
...
+    
+    @patch("keyring.get_password", side_effect=keyring_get_password)
+    def test_profile_loading_with_valid_schema(self, get_pass_func):
+        """
+        Test Validation, no error should be raised for valid schema
+        """
+        # Setup - copy profile to fake filesystem created by pyfakefs
+        custom_file_path = os.path.join(self.custom_dir, "zowe.config.json")
+        shutil.copy(self.original_nested_file_path, custom_file_path)
+        custom_file_path = os.path.join(self.custom_dir, "zowe.schema.json")
```
aadityasinha-dotcom, line 396: @t1m0thyj I have added the test but the fake file system is not taking the `zowe.schema.json` file

src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -12,9 +12,12 @@
 
 import commentjson
 from jsonschema import validate
+import os
+from typing import Union
+import requests
 
 
-def validate_config_json(path_config_json: str, path_schema_json: str):
+def validate_config_json(path_config_json: Union[str, dict], path_schema_json: str):
```
zFernand0, line 20: Is it ok to make the paht_schema_json: str optional?
And maybe we just pull it from the config file?

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -120,17 +131,51 @@ def init_from_file(self) -> None:
...
+        if path_schema_json and opt_out:
+            validate_config_json(path_config_json, path_schema_json)
+        
+
     def get_profile(
         self,
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
+        config_type: Optional[str] = None,
+        opt_out: Optional[bool] = True,
```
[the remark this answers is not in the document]
t1m0thyj: I agree we should make the names clear as to what is being opted in/out of 👍 
Perhaps `opt_in` and `opt_out` are unnecessary in the option name.
Could we have `validate_schema=True` and `override_with_env=False`?

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -129,7 +131,27 @@ def get_profile(
...
             cfg_profile = cfg.get_profile(
-                profile_name=profile_name, profile_type=profile_type
+                profile_name=profile_name, profile_type=profile_type, config_type=config_type, opt_out=opt_out
+            )
+        except jsonschema.exceptions.ValidationError as exc:
+            raise jsonschema.exceptions.ValidationError(
+                f"Instance was invalid under the provided $schema property, {exc}"
+            )
+        except jsonschema.exceptions.SchemaError as exc:
+            raise jsonschema.exception.SchemaError(
```
zFernand0, line 141: `jsonschema.exceptions`

zFernand0 (changes requested):

tests/unit/test_zowe_core.py
```diff
@@ -380,6 +388,49 @@ def test_profile_not_found_warning(self, get_pass_func):
...
+        props: dict = prof_manager.load(profile_name="zosmf")
+    
+    @patch("keyring.get_password", side_effect=keyring_get_password)
+    def test_profile_loading_with_invalid_schema(self, get_pass_func):
+        """
+        Test Validation, no error should be raised for valid schema
+        """
+        # Setup - copy profile to fake filesystem created by pyfakefs
+        with self.assertRaises(ValidationError):
+            custom_file_path = os.path.join(self.custom_dir, "invalid.zowe.config.json")
```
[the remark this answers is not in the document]
t1m0thyj: > Note: You may want to be cautious when selecting which files to copy, because the test can get confusing very quickly 😋

Should we generate invalid JSON files programmatically to place in the fake file system? For example, we could load a valid JSON file from disk, change one of the properties to make it invalid, and then save the invalid JSON to the fake filesystem. I think this would reduce the number of JSON files needed in the fixtures folder and hopefully make the tests a bit less confusing 🙂

tests/unit/test_zowe_core.py
```diff
@@ -384,6 +399,71 @@ def test_profile_not_found_warning(self, get_pass_func):
...
+                "profiles.zosmf.properties.password": "password",
+            })
+
+            # Test
+            prof_manager = ProfileManager(appname="invalid.zowe")
+            prof_manager.config_dir = self.custom_dir
+            props: dict = prof_manager.load(profile_name="zosmf", validate_schema=True)
+
+    @patch("keyring.get_password", side_effect=keyring_get_password)
+    def test_profile_loading_with_invalid_schema_internet_URI(self, get_pass_func):
```
aadityasinha-dotcom, line 447: Added test for `SchemaError` with `InternetURI`

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -174,7 +177,27 @@ def get_profile(
...
+                f"A type checker was asked to check a type it did not have registered, {exc}"
+            )
+        except jsonschema.exceptions.UnknownType as exc:
+            raise jsonschema.exceptions.UnknownType(
+                f"Unknown type is found in schema_json, exc"
+            )
+        except jsonschema.exceptions.FormatError as exc:
+            raise jsonschema.exceptions.FormatError(
+                f"Validating a format config_json failed for schema_json, {exc}"
             )
```
t1m0thyj, line 201: I think the exception handling should be moved to the `validate_schema` method in `config_file.py`. This would ensure consist error messages whether the schema is validated through the ConfigFile or ProfileManager class.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -116,7 +118,12 @@ def location(self, dirname: str) -> None:
         else:
             raise FileNotFoundError(f"given path {dirname} is not valid")
 
-    def init_from_file(self) -> None:
+    def init_from_file(
+        self, 
+        config_type: str, 
+        validate_schema: Optional[bool] = False,
```
t1m0thyj, line 124: I think the default value for `validate_schema` should be True 😋

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -128,14 +135,46 @@ def init_from_file(self) -> None:
...
+            self.validate_schema(config_type, validate_schema, verify=verify)
         # loading secure props is done in load_profile_properties
         # since we want to try loading secure properties only when
         # we know that the profile has saved properties
         # self.load_secure_props()
 
+    def validate_schema(
+        self,
+        config_type: str,
+        validate_schema: Optional[bool] = False,
```
t1m0thyj, line 152: Is the `validate_schema` option required on this function? If the`validate_schema`method is called, I would expect it to always validate the schema.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -184,6 +223,9 @@ def get_profile(
         self,
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
+        config_type: Optional[str] = None,
+        validate_schema: Optional[bool] = False,
```
t1m0thyj, line 227: I think the default value for `validate_schema` should be True (same as above) 😋

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -207,20 +230,23 @@ def get_profile(
                     SecurePropsNotFoundWarning,
                 )
         except Exception as exc:
+            print("hi", exc)
```
t1m0thyj, line 233: Can this be removed? Looks like it was added for debug purposes 🙂

t1m0thyj (changes requested):

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -128,14 +135,46 @@ def init_from_file(self) -> None:
             profile_jsonc = commentjson.load(fileobj)
```
t1m0thyj, line 135: If we assign to `self.profile_jsonc` instead, we will have the entire config file stored in memory as a class variable.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -128,14 +135,46 @@ def init_from_file(self) -> None:
...
+
+        path_schema_json = self.schema_path
+        if path_schema_json is None:    # check if the $schema property is not defined
+            warnings.warn(
+                f"$schema property could not found"
+            )
+                
+        # validate the $schema property 
+        if path_schema_json and validate_schema:
+            validate_config_json(path_config_json, path_schema_json, cwd = self.location, verify=verify)
```
t1m0thyj, line 176: Passing `path_config_json` as a string requires the file to be loaded from disk a 2nd time even though it has already been loaded into memory. Also sometimes the file may be loaded from the wrong directory.

Could we pass the contents of `zowe.config.json` as a JSON object instead? The JSON object could be stored on the class as `self.profile_jsonc` (see comment above)

src/core/zowe/core_for_zowe_sdk/config_file.py
aadityasinha-dotcom: Setting this to True, cause failing some unit tests. That's why I set it to False. 
I will make some changes to those tests :+1:

src/core/zowe/core_for_zowe_sdk/profile_manager.py
aadityasinha-dotcom: I tried moving it to the `config.py` but it doesn't raise error it skips the profile.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -128,14 +135,45 @@ def init_from_file(self) -> None:
...
+        if self.schema_property and validate_schema:
+            self.validate_schema(config_type, validate_schema)
         # loading secure props is done in load_profile_properties
         # since we want to try loading secure properties only when
         # we know that the profile has saved properties
         # self.load_secure_props()
 
+    def validate_schema(
+        self,
+        config_type: str,
```
t1m0thyj, line 151: Could you remove the `config_type` parameter since it's no longer used in this function?

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -128,14 +135,45 @@ def init_from_file(self) -> None:
...
+
+        Returns
+        -------
+        file_path to the $schema property
+        """
+
+        path_schema_json, path_config_json = None, ""
+
+        if self._location:
+            path_config_json = f"./{self.filename}"
```
t1m0thyj, line 165: Please remove the `path_config_json` variable since it's no longer used

tests/unit/test_zowe_core.py
```diff
@@ -170,17 +170,32 @@ def setUp(self):
...
+        )
         self.fs.add_real_file(self.original_file_path)
         self.fs.add_real_file(self.original_user_file_path)
         self.fs.add_real_file(self.original_nested_file_path)
         self.fs.add_real_file(self.original_schema_file_path)
-
+        self.fs.add_real_file(self.original_invalid_file_path)
+        self.fs.add_real_file(self.original_invalid_schema_file_path)
+        self.fs.add_real_file(self.original_invalidUri_file_path)
+        self.fs.add_real_file(self.original_invalidUri_schema_file_path)
```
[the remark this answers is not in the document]
t1m0thyj: Created #209 for this, no need to be concerned about it for this PR 🙂

t1m0thyj (commented):
Thanks for addressing earlier comments, validating schema works for me with a global team config file in home directory now 👍 

Left a few comments on config_file.py. Once those are addressed I'll approve since coverage looks good as well 🙂

tests/unit/test_zowe_core.py
```diff
@@ -166,22 +166,38 @@ class TestZosmfProfileManager(TestCase):
     def setUp(self):
         """Setup fixtures for ZosmfProfile class."""
         # setup pyfakefs
+        self.session_arguments = {"verify": False}
```
aadityasinha-dotcom, line 169: @t1m0thyj I did changed the verify flag to `false` for verification, please review that :relaxed:

src/core/zowe/core_for_zowe_sdk/validators.py
```diff
@@ -12,9 +12,12 @@
 
 import commentjson
 from jsonschema import validate
+import os
+from typing import Union, Optional
+import requests
 
 
-def validate_config_json(path_config_json: str, path_schema_json: str):
+def validate_config_json(path_config_json: Union[str, dict], path_schema_json: str, cwd: str):
```
zFernand0, line 20: curious if we should make `cwd` optional and fallback to `$PWD` 😋 

Either way, this new option need to be added as part of the parameter list below 😋

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -212,14 +234,15 @@ def get_profile(
...
-            return cfg_profile
+        
+        return cfg_profile
 
     def load(
         self,
         profile_name: Optional[str] = None,
         profile_type: Optional[str] = None,
         check_missing_props: bool = True,
+        validate_schema: Optional[bool] = True,
```
zFernand0, line 245: should we add the parameter list to the description below?
I think these are self-descriptive parameters, so maybe there is no need to do this.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -108,7 +111,11 @@ def location(self, dirname: str) -> None:
         else:
             raise FileNotFoundError(f"given path {dirname} is not valid")
 
-    def init_from_file(self) -> None:
+    def init_from_file(
+        self, 
+        config_type: str,
```
zFernand0, line 116: it would seem like this config_type is never used.
is it really needed?

And if we are removing it from here, then maybe we remove it from the `get_profile` method as well?

aadityasinha-dotcom:
> changes LGTM! left a few comments, not a huge deal though 😋
> 
> still going through the tests to see what's happening 😋

I tested this locally it was working fine though :thinking:

src/core/zowe/core_for_zowe_sdk/config_file.py
aadityasinha-dotcom: Yes, that makes sense :blush:

zFernand0 (approved):
LGTM! :yum:

t1m0thyj (approved):

Issue #194, opened 2023-07-06
Run SDK build workflow in multiple environments


t1m0thyj:
For the [`sdk-build.yaml` workflow](https://github.com/zowe/zowe-client-python-sdk/blob/main/.github/workflows/sdk-build.yml), we should add a matrix to run the build across multiple OSes and Python versions:
* Windows, Linux, macOS
* Python 3.7, 3.8, 3.9, 3.10, 3.11

Aman123lug:
This issue is a good first issue I will start contributing from this.
Okay! I need a add build support for multiple OS with different python versions.

t1m0thyj:
Thanks, I've assigned the issue to you 🙂 

Here's an example of a multi-dimensional matrix for GH workflows that may be helpful: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-a-multi-dimension-matrix

Aman123lug:
can you please review my PR and if i did something let me know 
thanks for sharing the docs

PR #199, opened 2023-07-15
Add matrix to build workflow

Aman123lug:
**What It Does**

Added a matrix to run the build across multiple OSes and Python versions #194

### DIFF
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        pip install -r requirements.txt
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -10,9 +10,11 @@ jobs:
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+      
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -8,14 +8,18 @@ on: [push, pull_request]
 jobs:
   build:
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
 
     steps:
     - uses: actions/checkout@v3
-    - name: Set up Python 3.7
+    - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v4
       with:
-        python-version: 3.7
+        python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,3 +2,8 @@
 
 All notable changes to the Zowe Client Python SDK will be documented in this file.
 
+## Recent Changes
+
+Added : Run SDK build workflow in multiple environments
+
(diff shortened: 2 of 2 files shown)


samadpls:
Hello @Aman123lug,

I've noticed that you're facing some challenges with test failures. I want to offer some suggestions:

- Instead of creating a new pull request on every change, you can clone the repository and push the changes directly.
- The reason you keep encountering build failures is that the current build only tests Python changes and not the build changes you made in the pull request.
- Regarding the changelog, I don't think you need to update it. @t1m0thyj can better guide you on whether the changelog needs to be updated or not.

Aman123lug:
@samadpls 
Thanks for your suggestions
I am new to this project and I need guidance can you please review my PR and tell me what I did wrong
If changelog does not require any update I will remove

Thank You!

CHANGELOG.md
t1m0thyj, line 1: Thanks for updating the changelog 🙂  For this PR we don't need to update the changelog because we only do so when the changes affect end users (see [changelog update guidelines](https://github.com/zowe/zowe-cli/blob/master/CONTRIBUTING.md#changelog-update-guidelines)).

Could you please remove the changelog update, since workflow changes aren't something that SDK users need to know about?

.github/workflows/sdk-build.yml
```diff
@@ -8,14 +8,18 @@ on: [push, pull_request]
 jobs:
   build:
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
```
t1m0thyj, line 15: The Python versions need to be wrapped in quotes. Currently `3.10` is being treated as `3.1` since YAML thinks it is a floating point number that should be rounded.

.github/workflows/sdk-build.yml
```diff
@@ -8,14 +8,18 @@ on: [push, pull_request]
 jobs:
   build:
     if: (github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository) && !contains(github.event.head_commit.message, '[ci skip]')
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
```
t1m0thyj, line 13: Please add `fail-fast: false` to the matrix definition so that if one build in the matrix fails, the others continue running (see [GitHub Actions docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures)).

t1m0thyj (changes requested):
@Aman123lug Thanks for contributing to the Zowe Python SDK 🙂 Left a few comments

CHANGELOG.md
Aman123lug: Okay i am removing the changelog Thank you

.github/workflows/sdk-build.yml
Aman123lug: Thanks for the feedback 
let me know, now it is correct or not 😅

t1m0thyj:
On line 28 of the workflow: `if [ -f requirements.txt ]; then pip install -r requirements.txt; fi`

The if condition written in Bash is unnecessary and will fail on Windows. I think this line could be replaced with just the pip command: `pip install -r requirements.txt`

Also, the PyYAML dependency in requirements.txt needs to be updated for newer versions of Python. I think we want to upgrade from version 5.4.1 to the latest one (6.0.1).

With these two changes, hopefully all the builds will pass 😋

samadpls:
Hey @Aman123lug , It appears that the `chanalog.md` file has been removed. Consequently, if this pull request is merged, it will result in the deletion of the `chanalog.md` file from the main branch as well.

Aman123lug:
Okay, then would you like me to create a new PR?
I changed the SDK-build.yml and added `pip install -r requirements.txt`
I also changed the PyYAML version latest one `6.0.1`

Aman123lug:
> Hey @Aman123lug , It appears that the `chanalog.md` file has been removed. Consequently, if this pull request is merged, it will result in the deletion of the `chanalog.md` file from the main branch as well.

Okay, Thanks for the review.
What should I do? Create new PR or change in Existing PR

samadpls:
> > Hey @Aman123lug , It appears that the `chanalog.md` file has been removed. Consequently, if this pull request is merged, it will result in the deletion of the `chanalog.md` file from the main branch as well.
> 
> Okay, Thanks for the review.
> What should I do? Create new PR or change in Existing PR
> 

I apologize for the confusion. Instead of creating a new pull request, please add the same 'changelog.md' file from this repository to your forked repository. Once you've done that, you'll be good to proceed.

Aman123lug:
> > > Hey @Aman123lug , It appears that the `chanalog.md` file has been removed. Consequently, if this pull request is merged, it will result in the deletion of the `chanalog.md` file from the main branch as well.
> > 
> > 
> > Okay, Thanks for the review.
> > What should I do? Create new PR or change in Existing PR
> 
> I apologize for the confusion😓. Instead of creating a new pull request, please add the same 'changelog.md' file from this repository to your forked repository. Once you've done that, you'll be good to proceed.

No worries. I created a new PR with all the same changes 😊

Issue #72, opened 2022-06-07
[core] Save secure profile properties to vault


t1m0thyj:
Given `zowe.config.json` and `zowe.config.user.json` files, store their secure properties in the vault (`Zowe/secure_config_props`) so credentials are remembered for future use.

PR #201, opened 2023-07-21
Save Profile Properties to Config File & Secure Vault Storage (#73, #72)

### LINKED ISSUE #73: [core] Save profile properties to zowe.config.json file
Given `zowe.config.json` and `zowe.config.user.json` files, create profiles in the highest priority layer, or update their properties in the layer where the profile exists.

samadpls:
**What It Does**

Resolves #73 , #72 

**How to Test**

**Review Checklist**
I certify that I have:
- [X] tested my changes
- [X] updated the changelog

**Additional Comments**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,4 +7,4 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 - Feature: Added a CredentialManager class to securely retrieve values from credentials and manage multiple credential entries on Windows [#134](https://github.com/zowe/zowe-client-python-sdk/issues/134)
 - Feature: Added method to load profile properties from environment variables
 - Feature: Added method to Save profile properties to zowe.config.json file [#73](https://github.com/zowe/zowe-client-python-sdk/issues/73)
-- Feature : Added method to Save secure profile properties to vault [#72](https://github.com/zowe/zowe-client-python-sdk/issues/72)
+- Feature: Added method to Save secure profile properties to vault [#72](https://github.com/zowe/zowe-client-python-sdk/issues/72)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,4 +6,5 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 - Feature: Added a CredentialManager class to securely retrieve values from credentials and manage multiple credential entries on Windows [#134](https://github.com/zowe/zowe-client-python-sdk/issues/134)
 - Feature: Added method to load profile properties from environment variables
-- Feature: Added method to Save profile properties to zowe.config.json file
+- Feature: Added method to Save profile properties to zowe.config.json file [#73](https://github.com/zowe/zowe-client-python-sdk/issues/73)
+- Feature : Added method to Save secure profile properties to vault [#72](https://github.com/zowe/zowe-client-python-sdk/issues/72)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,3 +2,6 @@
 
 All notable changes to the Zowe Client Python SDK will be documented in this file.
 
+## Recent Changes
+
+- Feature: Added method to Save profile properties to zowe.config.json file
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -464,7 +464,7 @@ class ConfigFile:
         # self.save("secure" in profile_data)
         
 
-    def save(self, secure_props=False):
+    def save(self, secure_props=True):
         """
         Save the config file to disk. and secure props to vault
         parameters:
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -374,38 +374,45 @@ class ConfigFile:
         is_property_secure = self.__is_secure(json_path)
         is_secure = secure if secure is not None else is_property_secure
         segments = json_path.split(".")[1:]
-        print(segments,'seg')
         updated_profiles = self.profiles
         prof_name = segments[0]
+
         print(updated_profiles,"updated_profiles")
         while len(segments) > 1:
             profile_name = segments[0]
             updated_profiles[profile_name] = updated_profiles.get(profile_name, {})
             # If profile name is the same as the one we are looking for and it is secure, add propeties to secure array
-            if prof_name == profile_name and is_secure:
-                updated_profiles["secure"] = updated_profiles.get("secure", [])
-                property_name = segments[0]
-                if property_name not in updated_profiles["secure"]:
-                    updated_profiles["secure"].append(property_name)
-            else:
-              updated_profiles = updated_profiles[profile_name]
-              segments.pop(0)
+            if prof_name == profile_name:
+                updated_profiles[profile_name]["secure"] = updated_profiles[profile_name].get("secure", [])
+                property_name = segments[-1]
+                if is_secure:
+                    if property_name not in updated_profiles[profile_name]["secure"]:
+                        updated_profiles[profile_name]["secure"].append(property_name)
+                    # Remove the property from properties dictionary if it's secure
+                    properties = updated_profiles[profile_name].get("properties")
+                    if properties and property_name in properties:
+                        del properties[property_name]
+                    
+                    if property_name not in updated_profiles[profile_name]["secure"]:
+                        updated_profiles[profile_name]["secure"].append(property_name)
+                    print(updated_profiles,"here")
+                else:
+                    if property_name in updated_profiles[profile_name]["secure"] and is_secure is not None:
+                        updated_profiles[profile_name]["secure"].remove(property_name)   
+                        # Set the property value in the properties dictionary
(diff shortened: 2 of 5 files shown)


src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -340,3 +340,60 @@ def load_secure_props(self) -> None:
...
+        Returns:
+            bool: True if the property should be stored securely, False otherwise.
+        """
+        profile_name, property_name = json_path.split(".")[1:]
+        profile = self.find_profile(profile_name, self.profiles)
+        if profile and profile.get("secure"):
+            return property_name in profile["secure"]
+        return False
+
+    def set_property(self, json_path, value, secure=False):
```
t1m0thyj, line 360: In order to match the behavior of the Node.js SDK, I think the default value of `secure` should be None.

* If `secure` is True, then we want to add the property name to the `secure` array and store the property value in the vault.
* If `secure` is False, then we want to remove the property name from the `secure` array and store the property value in plain text in the JSON file.
* If `secure` is None, then we want to detect whether the property is in the `secure` array. If so we treat it as secure, and if not we treat it as plain text.

t1m0thyj (commented):
Looks like a good start to me 👍 Left one comment

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -320,22 +320,21 @@ def load_profile_properties(self, profile_name: str) -> dict:
...
             lst.pop()
 
-
         # load secure props only if there are secure fields
         if secure_fields:
             CredentialManager.load_secure_props()
             self.secure_props=CredentialManager.secure_props.get(self.filepath, {})
             # load properties with key as profile.{profile_name}.properties.{*}
-            for (key, value) in self.secure_props.items():
+            for (key, value) in CredentialManager.secure_props.items():
```
t1m0thyj, line 337: I think we still want to use `self.secure_props` here, because it is an object containing secure properties for a config file where the keys are JSON paths and the values are secure credentials.

For the `CredentialManager.secure_props` object, the keys are file paths, and the values are an object structured like `self.secure_props`. The regex for matching JSON paths won't work on file paths.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -348,3 +347,145 @@ def load_profile_properties(self, profile_name: str) -> dict:
...
+            existing_secure_fields = current_profile.get("secure", [])
+            new_secure_fields = [field for field in secure_fields if field not in existing_secure_fields]
+
+            # JSON paths for new secure properties and store their values in CredentialManager.secure_props
+            for field in new_secure_fields:
+                json_path = f"{profile_path}.properties.{field}.value"
+                CredentialManager.secure_props[self.filepath] = {
+                    **CredentialManager.secure_props.get(self.filepath, {}),
+                    json_path: profile_data["properties"][field]
+                }
```
t1m0thyj, line 444: When `set_profile` is called I think we want to overwrite secure props if they already exist for the profile.

Before the for loop, we could reset the object to be empty:
`CredentialManager.secure_props[self.filepath] = {}`
And inside the for loop, we could eliminate the spread operator:
`CredentialManager.secure_props[self.filepath][json_path] = profile_data["properties"][field]`

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -348,3 +347,145 @@ def load_profile_properties(self, profile_name: str) -> dict:
...
+        """
+        Save the config file to disk. and secure props to vault
+        parameters:
+            secure_props (bool): If True, the secure properties will be stored in the vault. Default is False.
+        Returns:
+            None
+        """
+        # Update the config file with any changes
+        if self.profiles is None:
+            self.init_from_file()
```
t1m0thyj, line 467: If `self.profiles` is None, perhaps we should throw an error here

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -348,3 +347,145 @@ def load_profile_properties(self, profile_name: str) -> dict:
...
+        elif any(self.profiles.values()):
+            with open(self.filepath, 'w') as file:
+                # Update the profiles in the JSON data
+                self.jsonc["profiles"] = self.profiles
+                file.seek(0)  # Move the file pointer to the beginning of the file
+                commentjson.dump(self.jsonc, file, indent=4)
+                file.truncate()  # Truncate the file to the current file pointer position
```
t1m0thyj, line 474: Are the `seek` and `truncate` calls necessary?

I believe when a file is opened in "w" mode any contents are deleted so the pointer is already at 0.

t1m0thyj (changes requested):
Left a few comments 🙂 Also please address failing tests

t1m0thyj (changes requested):
I'm having some issues when testing this PR using the following procedure:
1. Using Zowe CLI, run `zowe config init --no-prompt` to create a zowe.config.json template.
2. Using the Python SDK, run a test script to update a profile property:
    ```
    from zowe.core_for_zowe_sdk import ProfileManager
    pm = ProfileManager(show_warnings=False)
    pm.load(profile_type="zosmf")
    pm.set_property("profiles.zosmf.properties.port", 1443)
    pm.save()
    ```
3. Receive an error because "zowe.config.user.json" is missing:
```
Traceback (most recent call last):
  File "/Users/Timothy/Projects/zowe/zowe-client-python-sdk/test.py", line 9, in <module>
    pm.set_property("profiles.zosmf.properties.port", 1443)
  File "/Users/Timothy/Projects/zowe/zowe-client-python-sdk/src/core/zowe/core_for_zowe_sdk/profile_manager.py", line 362, in set_property
    highest_priority_layer = self.get_highest_priority_layer(json_path)
  File "/Users/Timothy/Projects/zowe/zowe-client-python-sdk/src/core/zowe/core_for_zowe_sdk/profile_manager.py", line 334, in get_highest_priority_layer
    layer.init_from_file()
  File "/Users/Timothy/Projects/zowe/zowe-client-python-sdk/src/core/zowe/core_for_zowe_sdk/config_file.py", line 119, in init_from_file
    self.autodiscover_config_dir()
  File "/Users/Timothy/Projects/zowe/zowe-client-python-sdk/src/core/zowe/core_for_zowe_sdk/config_file.py", line 234, in autodiscover_config_dir
    raise FileNotFoundError(f"Could not find the file {self.filename}")
FileNotFoundError: Could not find the file zowe.config.user.json
```

CHANGELOG.md
```diff
@@ -5,4 +5,6 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 ## Recent Changes
 
 - Feature: Added a CredentialManager class to securely retrieve values from credentials and manage multiple credential entries on Windows [#134](https://github.com/zowe/zowe-client-python-sdk/issues/134)
-- Feature: Added method to load profile properties from environment variables
\ No newline at end of file
+- Feature: Added method to load profile properties from environment variables
+- Feature: Added method to Save profile properties to zowe.config.json file [#73](https://github.com/zowe/zowe-client-python-sdk/issues/73)
+- Feature : Added method to Save secure profile properties to vault [#72](https://github.com/zowe/zowe-client-python-sdk/issues/72)
```
traeok, line 10: looks like there's an extra space on line 10 🙂

src/core/zowe/core_for_zowe_sdk/config_file.py
traeok, line 1: @t1m0thyj I don't see any issues with formatting here, but curious: would the Python SDK benefit from a linter such as `black`? I see that `flake8` is installed, but `black` would allow us to apply more opinionated styling, similar to our prettier usage in other repos.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -42,6 +42,8 @@
 
 # Profile datatype is used by ConfigFile to return Profile Data along with
 # metadata such as profile_name and secure_props_not_found
+
+
 class Profile(NamedTuple):
```
traeok, line 47: For the `autodiscover_config_dir` function, do we want to skip over `zowe.config.user.json` if it does not exist? It is optional, so we should avoid throwing exceptions for a missing user-level config.

traeok (changes requested):
Nice work thus far @samadpls!

Regarding the error w/ a missing `zowe.config.user.json`:

If I provide a `zowe.config.user.json` in my working directory, the code continues, but it then tries to open a `zowe.config.user.json` in my Zowe home directory. It fails here as it cannot find the file in my `.zowe` folder.

However, the code looks good so far, so I think once the error handling is added for accessing `zowe.config.user.json`, it should work as intended 🙂

CHANGELOG.md
samadpls: Yes, the text is on the same line, but displayed on a new line on GitHub.

src/core/zowe/core_for_zowe_sdk/config_file.py
samadpls: This could be resolved by implementing a try-except block while invoking the function.

traeok (approved):
Tested w/ changes and LGTM, thanks @samadpls!

t1m0thyj (approved):
LGTM, thanks @samadpls!

PR #203, opened 2023-07-31
Fixes profile merge order to match Node.js SDK

### LINKED ISSUE #190: Fix profile merge order to match Node.js SDK
Tested https://github.com/zowe/zowe-client-python-sdk/pull/144 with the [sample config and profiles](https://github.com/zowe/zowe-client-python-sdk/files/10864029/zowe-client-python-sdk-test.zip) that we've been using. For the `my_zosmf` profile it is failing to load user and password:
Python SDK: `{'host': 'example.com', 'rejectUnauthorized': True, 'port': 443}`
Node.js SDK: `{"port":443,"host":"example.com","rejectUnauthorized":true,"user":"admin","password":"password"}`

The `ProfileManager.load` method loops through config files in the following order:
1. Project User Config (`./zowe.config.user.json`)
2. Project Config (`./zowe.config.json`)
3. Global User Config (`~/zowe.config.user.json`)
4. Global Config (`~/zowe.config.json`)

In the current implementation, profile properties are first loaded from each layer of the config individually, and merged together afterwards.

For the case of our sample config:
* The `user` and `password` properties are stored in the `my_base` profile of Project User Config (`./zowe.config.user.json`).
* This `my_base` profile is defined as the default base profile in Project Config (`./zowe.config.json`).
* At the time when we load Project User Config, we do not yet know that `my_base` is the default base profile, so the current logic will skip loading its properties. 😢 

To fix this, I believe refactoring `ProfileManager.load` will be necessary. We should first merge all the config layers together, and then load profile properties as a single operation from the resulting merged object.

For a detailed explanation of how config layers should be merged together, see [Zowe docs](https://docs.zowe.org/stable/user-guide/cli-using-understand-profiles-configs).

_Originally posted by @t1m0thyj in https://github.com/zowe/zowe-client-python-sdk/pull/144#discussion_r1185573369_

aadityasinha-dotcom:
**What It Does**

Resolves #190 

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/profile_manager.py
+++ b/src/core/zowe/core_for_zowe_sdk/profile_manager.py
@@ -162,7 +162,6 @@ class ProfileManager:
         cfg: ConfigFile,
         profile_name: Optional[str],
         profile_type: Optional[str],
-        config_type: Optional[str],
         validate_schema: Optional[bool] = True,
     ) -> Profile:
         """
@@ -232,7 +231,7 @@ class ProfileManager:
                 )
         except Exception as exc:
             warnings.warn(
-                f"Could not load {config_type} '{cfg.filename}' at '{cfg.filepath}'"
+                f"Could not load '{cfg.filename}' at '{cfg.filepath}'"
                 f"because {type(exc).__name__}'{exc}'.",
                 ConfigNotFoundWarning,
             )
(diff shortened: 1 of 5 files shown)


src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -259,10 +260,17 @@ def load(
 
         missing_secure_props = []  # track which secure props were not loaded
 
+        loaded_cfg: dict = {}
+
         for i, (config_type, cfg) in enumerate(config_layers.items()):
+            loaded_cfg = always_merger.merge(loaded_cfg, cfg.profiles)
```
aadityasinha-dotcom, line 266: I did used the `deepmerge` library so that it can merge each layer as a whole `config`

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -259,10 +260,17 @@ def load(
...
+            if loaded_cfg:
+                cfg.profiles = loaded_cfg
+                
             profile_loaded = self.get_profile(
```
t1m0thyj, line 270: This `get_profile` call and the lines that follow it could be moved outside of the for loop. We only need to load profile properties, secure properties, and environment variables once after we have the merged config object.

src/core/zowe/core_for_zowe_sdk/profile_manager.py
t1m0thyj: Not all config layers can be merged with the same merger function. See the [Node.js SDK](https://github.com/zowe/imperative/blob/edd7bf378b2007416b1824ea05db93d32a248368/packages/config/src/Config.ts#L552-L569) for how profiles of each layer should be merged:
1. Project and Project User are deep merged into a temporary Project Merged layer
2. Global and Global User are deep merged into a temporary Global Merged layer
3. Add profiles to the Project Merged layer which are only present in Global Merged layer

t1m0thyj (changes requested):

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -273,12 +278,21 @@ def load(
...
+        project = self.project_config.profiles
+        prjt = always_merger.merge(usrProject, project)
+
+        usrGlobal = self.global_user_config.profiles
+        glbal = self.global_config.profiles
+        glbl = always_merger.merge(usrGlobal, glbal)
+
+        if override_with_env:
+                env_var = {**self.get_env(cfg)}
+
```
aadityasinha-dotcom, line 295: @t1m0thyj need some reviews on it :relaxed:

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -273,12 +278,21 @@ def load(
...
             if i == 1 and profile_props:
                 break  # Skip loading from global config if profile was found in project config
 
+        usrProject = self.project_user_config.profiles
+        project = self.project_config.profiles
+        prjt = always_merger.merge(usrProject, project)
+
+        usrGlobal = self.global_user_config.profiles
+        glbal = self.global_config.profiles
+        glbl = always_merger.merge(usrGlobal, glbal)
```
t1m0thyj, line 291: The `merge` function modifies the contents of the first object passed in. We don't want to be modifying the `profiles` property on config classes - it should match what was loaded from the zowe.config.json file.

I think we should use the `deepcopy` method from the Python `copy` module to create copies of `usrProject` and `usrGlobal` before performing the merge.

aadityasinha-dotcom: I tried using the `deepcopy` like this `usrGlobal = deepcopy(self.global_user_config.profiles)` but it doesn't work

aadityasinha-dotcom:
@t1m0thyj please review this PR :blush:

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -273,12 +279,29 @@ def load(
...
+        usrProjectCopy = deepcopy(usrProject)
+        projectCopy = deepcopy(project)
+        prjt = always_merger.merge(usrProject, project)
+
+        usrGlobal = self.global_user_config.profiles
+        glbal = self.global_config.profiles
+        
+        # Creating copies of the usrGlobal and glbal objects
+        usrGlobalCopy = deepcopy(usrGlobal)
+        glbalCopy = deepcopy(glbal)
```
aadityasinha-dotcom, line 299: I have created the copies using `deepcopy` for the objects before getting merged

zFernand0 (approved):
I'm having a tough time trying to find differences in behavior with the NodeJS SDK.
LGTM! 😋

traeok (approved):

PR #204, opened 2023-08-01
chore: Update README with developer setup steps

traeok:
**What It Does**

This PR adds set-up instructions to the Python SDK for developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,12 +36,12 @@ pip install zowe.<subpackage>_for_zowe_sdk
 ```
 
 For more information on the available sub-packages click [HERE](https://zowe-client-python-sdk.readthedocs.io/en/latest/packages/packages.html)
-
-**Note**: If you want to stay on the bleeding edge, install the Zowe SDK from the `main` branch with the following command:
-
-```
-pip install zowe@git+https://github.com/zowe/zowe-client-python-sdk@main#subdirectory=src
-```
+<!-- 
+    **Note**: If you want to stay on the bleeding edge, install the Zowe SDK from the `main` branch with the following command:
+    ```
+    pip install zowe@git+https://github.com/zowe/zowe-client-python-sdk@main#subdirectory=src
+    ``` 
+-->
 
 ## Requirements
 
--- a/README.md
+++ b/README.md
@@ -12,16 +12,16 @@ that can interact with z/OS components.
 
 ## Installation
 
-
-When installing the Zowe Client Python SDK you have two options.
+When installing the Zowe Client Python SDK, you have two options:
 
 - Install all the Zowe packages
 - Install a single sub-package
 
-The choice depends on your intentions. If you chose to install all Zowe SDK packages
+The choice depends on your intentions. If you choose to install all Zowe SDK packages,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -54,6 +54,50 @@ pyyaml
...
 
+### Developer setup
+
+Clone the repository using `git`:
+```
+git clone https://github.com/zowe/zowe-client-python-sdk.git
+```
+
+We recommend that developers create a virtual environment to install all the required dependencies for this project.  
+**Note:** Use of `venv`, the virtual environment module, requires at least version 3.3 of Python.
```
t1m0thyj, line 65: The Zowe Python SDK itself requires Python 3.7 or newer (because it uses features added in 3.7 like dataclasses).

Would it make sense to add a note about this higher up in the readme, and then this one about venv would no longer be needed?

README.md
```diff
@@ -54,6 +54,50 @@ pyyaml
...
+
+To create a virtual environment, make a folder to store your virtual environment (referred to below as `YOUR_VENV_PATH`). Then, use the `venv` module to set it up:
+
+```
+python3 -m venv YOUR_VENV_PATH
```
t1m0thyj, line 70: Is `python3` available on all systems? I thought on Windows the binary was just `python` but am probably remembering wrong 😋

README.md
```diff
@@ -54,6 +54,50 @@ pyyaml
...
+
+Next, activate your virtual environment so that Python uses this folder to manage dependencies.
+
+Assuming that you are using Bash shell, reference the command below. Otherwise, check the table titled "Command to activate virtual environment" [here](https://docs.python.org/3/library/venv.html#how-venvs-work) to find the command that works for your shell.
+
+```
+source YOUR_VENV_PATH/bin/activate
+```
+
+Navigate to the root of the repository and checkout the desired branch. Currently, active development is on the `main` branch.
```
t1m0thyj, line 81: Should this step be moved above creating a virtual environment?

I think the `venv` is typically created in a subfolder of the repository (not required to do so, but probably recommended).

t1m0thyj: For people who are new to Python and may not know what to supply for `YOUR_VENV_PATH`, should we suggest a default of either `venv` or `.venv` in the repo folder?

t1m0thyj (commented):
Thanks for updating the readme, left a few comments 🙂 

Sorry if its a bit outside the scope of this PR, but would you mind reverting the following change in both `./requirements.txt` and `./docs/requirements.txt`? https://github.com/zowe/zowe-client-python-sdk/pull/189/commits/9a6efa4ac99ff314a565716bfd62beaa63f14838
When contributors install from `requirements.txt`, we want to link each SDK so that edits to source code take affect immediately without reinstalling 😋

README.md
traeok: I will update the instructions to create a `venv` folder in the repo folder, since its `gitignore`'d it should be fine 😋 

I think I'll go with `venv`

t1m0thyj (approved):
LGTM and thanks for including the "deactivate" command - I always forget about it 😋

PR #210, opened 2023-08-29
Reduce duplication of Codecov uploads

t1m0thyj:
Codecov was uploading coverage info for every OS and Python version which seems a bit excessive 😋 
Sometimes this was causing the build to fail.
This PR reduces the Codecov check to run only once per OS.

### DIFF
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -14,6 +14,9 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk-build.yml
```diff
@@ -39,8 +42,10 @@ jobs:
         coverage xml
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v3
+      if: ${{ matrix.python-version == '3.7' }}
```
zFernand0, line 45: should we pick the latest so we don't have to update it so often? 😋

zFernand0 (approved):
LGTM! 😋 
left a small comment, but no need to worry too much about it 😋

Issue #211, opened 2023-09-04
create_data_set + delete_data_set: datasets are not being created properly if there is a '#' symbol in their names


KUGDev:
**Describe the bug**

I've encountered an issue creating and deleting dataset with a '#' symbol in it's name
Example: MYNAME.TEST#ZWE

**Expected and actual results**

1. Expected: dataset MYNAME.TEST#ZWE is created and deleted by these functions
2. Actual: dataset is created as MYNAME.TEST#ZWE. If I try to delete the MYNAME.TEST#ZWE dataset, created manually, it's giving an error, as it tries to find MYNAME.TEST

**Describe your environment**

z/OS 2.5, zowe-client-python-sdk 1.0.0-dev.10

**Additional context**

adam-wolfe:
Implemented by PR: #212

PR #212, opened 2023-09-04
GH-211 Fixed issue for datasets with '#' in URL

KUGDev:
**What It Does**

fix for #211 

**How to Test**

Test all the functions, related to work with datasets and members with '#' character in their names

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/sdk_api.py
+++ b/src/core/zowe/core_for_zowe_sdk/sdk_api.py
@@ -10,6 +10,7 @@ SPDX-License-Identifier: EPL-2.0
 Copyright Contributors to the Zowe Project.
 """
 
+import urllib
 from .exceptions import UnsupportedAuthType
 from .request_handler import RequestHandler
 from .session import Session, ISession
@@ -60,3 +61,14 @@ class SdkApi:
         dictionary creation
         """
         return self.request_arguments.copy()
+
+    def _encode_uri_component(self, str_to_adjust):
+        """Adjust string to be correct in a URL
+
+        Returns
+        -------
+        adjusted_str
+            A string with special characters, acceptable for a URL
+        """
+
+        return urllib.parse.quote(str_to_adjust, safe="!~*'()") if str_to_adjust is not None else None
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
@@ -117,7 +117,7 @@ class Files(SdkApi):
             A JSON with a list of dataset names (and attributes if specified) matching the given pattern.
         """
         custom_args = self._create_custom_request_arguments()
-        custom_args["params"] = {"dslevel": name_pattern}
+        custom_args["params"] = {"dslevel": self._encode_uri_component(name_pattern)}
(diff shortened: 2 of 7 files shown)


src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -47,6 +47,18 @@ def __init__(self, connection):
...
+    def __adjust_for_url(self, str_to_adjust):
+        """Adjust string to be correct in a URL
+
+        Returns
+        -------
+        adjusted_str
+            A string with special characters, acceptable for a URL
+        """
+
+        return str_to_adjust.replace("#", "%23") if str_to_adjust is not None else None
```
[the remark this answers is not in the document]
KUGDev: @t1m0thyj thanks for the comment. I've tried other possible dataset names, looks like they are ok to use inside of URL (such as with @ and $ chars in them). But of course, it could be better to use the `urllib.parse.quote` to prevent further issues. Will do

KUGDev:
@t1m0thyj hi, I've added the changes. Should I do >80% coverage? Some of the functions are not covered with tests, so the codecov does not allow to pass the patch as successful

KUGDev:
Hi, I also discovered, that this problem relates to other places in the SDK
Will be working on this

KUGDev:
@t1m0thyj @adam-wolfe I've finished the issue, could you please make a review?

src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -60,3 +61,14 @@ def _create_custom_request_arguments(self):
...
+    def _adjust_for_url(self, str_to_adjust):
+        """Adjust string to be correct in a URL
+
+        Returns
+        -------
+        adjusted_str
+            A string with special characters, acceptable for a URL
+        """
+
+        return urllib.parse.quote(str_to_adjust) if str_to_adjust is not None else None
```
adam-wolfe, line 74: @t1m0thyj, correct me if I am wrong, but I think we may want to do something like:

`return urllib.parse.quote(str_to_adjust, safe="!~*'()") if str_to_adjust is not None else None`

Otherwise, it will encode '*' which causes issues with files.list_dsn. I am not sure what the z/OSMF APIs expect for the other characters ("!~'()"). I found this info in a Stack Overflow comment: https://stackoverflow.com/a/74439601

adam-wolfe (changes requested):

src/core/zowe/core_for_zowe_sdk/sdk_api.py
t1m0thyj: Good catch, we want to mimic the escaping behavior of the JS function `encodeURIComponent` used in the Zowe Node.js SDKs.

KUGDev:
Hello @t1m0thyj @adam-wolfe 
Thanks for the comments, I've changed the functionality according to the requirements

adam-wolfe (approved):

adam-wolfe:
Looks good to me. Thanks!

t1m0thyj (approved):
LGTM, thanks @KUGDev!

PR #245, opened 2023-12-22
return response instead of raw for stream requests

rupurt:
**What It Does**

- enables use of iterator
- https://requests.readthedocs.io/en/latest/user/quickstart/#raw-response-content

**How to Test**
- Download dataset via stream
```python
response = zosmf_files.get_dsn_content_streamed(dsn)
with open(destination_path, "wb") as fd:
    for chunk in response.iter_content(chunk_size=256):
        fd.write(chunk)
        size += len(chunk)
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 ### Bug Fixes
 
 - Fixed 'create_data_set' to accept "FBA", "FBM", "VBA", "VBM" as valid recfm [#240](https://github.com/zowe/zowe-client-python-sdk/issues/240)
+- Return response instead of raw from streamed requests
 
 ## `1.0.0-dev12`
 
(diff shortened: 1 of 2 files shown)


t1m0thyj (approved):
@rupurt Thanks for your contribution!

Although this is a breaking change for anyone using SDK methods that return streams, it seems to follow the best practices recommended by the `requests` package 👍

rupurt:
@t1m0thyj  yeah, that would be unfortunate. But I feel like this is something worth making a breaking change for. Currently you need to manually decode the gzipped responses. If we want to avoid a breaking change, how do you feel about an extra argument `raw = True` ?

zFernand0 (approved):
I agree with the position of just making the breaking change.
We do want to handle as much as it makes sense, and also follow recommended practices.

LGTM! 😋

PR #246, opened 2023-12-26
Fix/correlator issue 242

### LINKED ISSUE #242: [zos-jobs] Listing by correlator doesn't work
> on `list_jobs()` I tried to use the correlator and z/OSMF kept saying "invalid correlator"
> but it worked on `spool_files()` or whatever that interface is
> using the `job-correlator` property returned from submit

Thanks to @tucker01 for reporting this 🙂 

It seems like we are hardcoding "owner" to always be present on the payload which is probably why the correlator has no effect:
https://github.com/zowe/zowe-client-python-sdk/blob/8c9580aae179f4b0898b5032c4d63d08fd6b75e8/src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py#L220-L222

samadpls:
**What It Does**

fixed #242 
**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,9 +6,8 @@
         "editor.codeActionsOnSave": {
             "source.organizeImports": true
         }
-    }
-  },
-  "python.testing.pytestArgs": ["tests"],
-  "python.testing.unittestEnabled": false,
-  "python.testing.pytestEnabled": true
+    },
+    "python.testing.pytestArgs": ["tests"],
+    "python.testing.unittestEnabled": false,
+    "python.testing.pytestEnabled": true
 }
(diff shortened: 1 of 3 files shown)


zFernand0 (commented):
This is a bit hard to test personally since no matter the correlator, I always get no jobs back.
I do see how not using the submitting user as the default owner might help though.

@tucker01, could you help verify that this quick fix works for you?

t1m0thyj (approved):
LGTM, thanks @samadpls! This matches the behavior of the Node.js SDK when no owner is passed in.

zFernand0 (approved):

PR #256, opened 2024-03-13
Fix missing GitHub releases

t1m0thyj:
Recent releases have tags created, but are missing a GitHub release:
image

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -16,7 +16,9 @@ module.exports = {
     plugins: [
         "@octorelease/changelog",
         "@octorelease/pypi",
-        "@octorelease/github",
+        ["@octorelease/github", {
+            publishRelease: true
+        }],
         "@octorelease/git"
     ]
 };
--- a/release.config.js
+++ b/release.config.js
@@ -2,7 +2,8 @@ module.exports = {
     branches: [
         {
             name: "main",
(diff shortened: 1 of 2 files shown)


zFernand0 (dismissed):

release.config.js
```diff
@@ -2,7 +2,8 @@ module.exports = {
     branches: [
         {
             name: "main",
-            level: "minor"
+            level: "none",
+            prerelease: true
```
zFernand0, line 6: do we need to specify the `channel`?
otherwise it might default to `main` and start tagging things as `1.0.0-main15, -main16...`

or will it remember to use the word `dev` for the prerelease strings?

t1m0thyj: This shouldn't matter because currently Octorelease isn't smart enough to handle versioning for Python projects, so we do it manually and pass the new version to Octorelease. The `level` and `prerelease` properties are still useful for Octorelease to validate the semver diff level and create the GitHub release with prerelease status.

zFernand0: Thanks for explaining. I guess that if Octorelease was able to handle versioning for Python projects, then we would've needed to specify which prerelease string to use (`dev`) somehow 😋

zFernand0 (approved):

PR #257, opened 2024-03-13
Fix error when issuing TSO commands

### LINKED ISSUE #255: Failed to list user details using send_tso_message() with the Zowe SDK version 1.0.0.dev10
Hello Team,

We are trying to execute a command to get user details from Mainframe using send_tso_message() method, and we're using the command "LU &lt;USERID&gt;" to get the information.

This command works fine on the version 0.2.0 available on pip, but if we try to execute the same command and method using the latest version v1.0.0.dev10 we are getting the error bellow:

```
IKJ56602I COMMAND SYSTEM RESTARTING DUE TO ERROR
READY
```

We are using the same script, for both SDK versions, with the same methods, the only difference is in the dictionary used for tso connection, but in general is all the same.

If you need more details or files, feels free to request.

We hope that you would help us to identity this issue.

Thanks is advance.

Regards

- Environment:
Python 3.8.17
pip 19.3.1 from /home/user/venv/.venv/lib64/python3.8/site-packages/pip (python 3.8)
NAME="Red Hat Enterprise Linux"
VERSION="8.9 (Ootpa)"

t1m0thyj:
Fixes #255 and updates the system test for `Tso.issue_command` to verify that the response is successful

### DIFF
--- a/tests/unit/test_zos_tso.py
+++ b/tests/unit/test_zos_tso.py
@@ -1,16 +1,16 @@
 """Unit tests for the Zowe Python SDK z/OS TSO package."""
 
-import unittest
+from unittest import TestCase, mock
 
 from zowe.zos_tso_for_zowe_sdk import Tso
 
 
-class TestTsoClass(unittest.TestCase):
+class TestTsoClass(TestCase):
     """Tso class unit tests."""
 
     def setUp(self):
         """Setup fixtures for Tso class."""
-        self.connection_dict = {
+        self.test_profile = {
             "host": "mock-url.com",
             "user": "Username",
             "password": "Password",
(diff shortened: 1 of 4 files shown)


tests/unit/test_zos_tso.py
```diff
@@ -20,5 +20,16 @@ def setUp(self):
 
     def test_object_should_be_instance_of_class(self):
         """Created object should be instance of Tso class."""
-        tso = Tso(self.connection_dict)
+        tso = Tso(self.test_profile)
         self.assertIsInstance(tso, Tso)
+
+    @mock.patch("requests.Session.send")
+    def test_issue_command(self, mock_send_request):
+        """Test issuing a command sends a request"""
```
zFernand0, line 28: Should we change the test description to something like `should ... `?

awharn (approved):

traeok (approved):
LGTM, thanks @t1m0thyj 😁

zFernand0 (commented):
LGTM! 😋

PR #268, opened 2024-05-23
Adding logger modules

pem70:
**What It Does**

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


.vscode/settings.json
```diff
@@ -9,5 +9,14 @@
...
-    "rust-analyzer.linkedProjects": ["./src/secrets/Cargo.toml"],
+    "rust-analyzer.linkedProjects": [
+        "./src/secrets/Cargo.toml"
+    ],
+    "python.analysis.extraPaths": [
+        "./src/core",
+        "./src/zos_console",
+        "./src/zos_files",
+        "./src/zos_jobs",
+        "./src/zosmf"
```
t1m0thyj, line 20: Thanks for adding this setting 🙂  Curious if it supports wildcards? If so it could be simplified to just one line: `./src/*`

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the Zowe Client Python SDK will be documented in this file.
 
+## Recent Changes
+
+### Enhancements
+
+- Added logger class to core SDK [#185] https://github.com/zowe/zowe-client-python-sdk/issues/185
```
t1m0thyj, line 9: For the issue number to be a Markdown link, it should be formatted like this:
```suggestion
- Added logger class to core SDK [#185](https://github.com/zowe/zowe-client-python-sdk/issues/185)
```

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -118,7 +120,9 @@ def init_from_file(
                 pass
 
         if self.filepath is None or not os.path.isfile(self.filepath):
-            warnings.warn(f"Config file does not exist at {self.filepath}")
+            if not suppress_config_file_warnings:
```
t1m0thyj, line 123: Now that we have logging in place, we might be able to remove the `suppress_config_file_warnings` flag and log some of the config file warnings at INFO level instead.

However I think that removing warnings that were already present would be outside the scope of the effort to add a logger, so this could be addressed in a separate issue 🙂

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -0,0 +1,21 @@
+import logging
+import os
+
+class Log:
+    """root logger setup and a funtion to customize logger level"""
+
+    dirname = os.path.join(os.path.expanduser("~"), ".zowe/logs")
```
t1m0thyj, line 7: The default location of the `.zowe` folder is `~/.zowe` as used here. In Zowe CLI and Zowe Explorer, it can be overridden with the `ZOWE_CLI_HOME` environment variable to point to another directory.

We may want to check for the existence of `os.environ["ZOWE_CLI_HOME"]`, and if it is defined then store the `logs` directory inside it. This could be addressed in a separate issue 🙂

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -0,0 +1,21 @@
...
+
+    if not os.path.isdir(dirname):
+        os.makedirs(dirname)
```
t1m0thyj, line 10: This could be optimized to just one line:
```suggestion
    os.makedirs(dirname, exist_ok=True)
```

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -0,0 +1,21 @@
...
+
+    logging.basicConfig(
+        filename=os.path.join(dirname, "python_sdk_logs.log"),
+        level=logging.INFO,
+        format="[%(asctime)s] [%(levelname)s] [%(name)s] - %(message)s",
+        datefmt="%m/%d/%Y %I:%M:%S %p",
+    )
```
t1m0thyj, line 17: If all the logs are written to the same file, over a long period of time it will grow in size and become many MB. We may want to investigate the `RotatingFileHandler` class which switches logging to a new file when the file exceeds a certain size, and deletes old files (similar to what our Node.js SDK does).

Since logging to a file is working, this could be addressed in a separate issue 🙂

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @staticmethod
+    def setLoggerLevel(level: int):
+        logging.root.setLevel(level)
```
t1m0thyj, line 21: I don't know if setting the level of `logging.root` is best practice here. If I understand correctly this will affect the level of all loggers, including ones that other Python packages besides the Zowe SDK may be using.

Since we create multiple loggers in each module that each have their own name, if we want to set logger level across all of them without affecting the root logger, we may need to track them in this class. Something like the following:
```suggestion
    loggers = []
    @staticmethod
    def registerLogger(name: string):
        Log.loggers.append(logging.getLogger(name))

    @staticmethod
    def setLoggerLevel(level: int):
        for logger in Log.loggers:
            logger.setLevel(level)
```

src/core/zowe/core_for_zowe_sdk/request_handler.py
```diff
@@ -66,35 +73,14 @@ def perform_request(self, method, request_arguments, expected_code=[200]):
...
+        self.__logger.debug(f"Request method: {self.method}, Request arguments: {self.request_arguments}, Expected code: {expected_code}")
         self.__validate_method()
-        self.__send_request()
+        self.__send_request(stream = stream)
         self.__validate_response()
+        if stream:
+            return self.response
         return self.__normalize_response()
 
-    def perform_streamed_request(self, method, request_arguments, expected_code=[200]):
```
t1m0thyj, line 74: Thanks for reducing code duplication here 😋

t1m0thyj (changes requested):
Looks pretty good, thanks @pem70 for your contribution! Left a few comments.

src/core/zowe/core_for_zowe_sdk/config_file.py
pem70: I agree with you. It should be addressed in a separate issue.

.vscode/settings.json
pem70: I don't think it supports wildcard as I tried it on VS Code.

src/core/zowe/core_for_zowe_sdk/logger.py
pem70: I think it could be addressed in a separate issue, too.

src/core/zowe/core_for_zowe_sdk/logger.py
pem70: It is a good idea. Maybe a separate issue?

pem70 (commented):
Thanks for reviewing! I don't find an urgent need for fixing these changes asap and they could be addressed in separate issues. However, if you think them should be done right now, I can still do it.

pem70 (commented):
Solve the bug in RegisterLogger(name: str)

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -213,6 +212,7 @@ def get_profile(
             self.init_from_file(validate_schema)
 
         if profile_name is None and profile_type is None:
+            self.__logger.error(f"Failed to load profile '{profile_name}' because Could not find profile as both profile_name and profile_type is not set")
```
traeok, line 215: Can we adjust this error for clarity? It looks like we are passing in the `profile_name` for the error that's logged, but `profile_name` will always be `None` if this condition is satisfied.

```suggestion
            self.__logger.error(f"Failed to load profile: profile_name and profile_type were not provided.")
```

src/core/zowe/core_for_zowe_sdk/credential_manager.py
```diff
@@ -49,6 +53,7 @@ def load_secure_props() -> None:
                 return
 
         except Exception as exc:
+            CredentialManager.__logger.error(f"Fail to load secure profile {constants['ZoweServiceName']}")
```
traeok, line 56:
```suggestion
            CredentialManager.__logger.error(f"Failed to load secure profile {constants['ZoweServiceName']}")
```

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -0,0 +1,26 @@
+import logging
+import os
+
+class Log:
+    """root logger setup and a funtion to customize logger level"""
```
traeok, line 5:
```suggestion
    """root logger setup and a function to customize logger level"""
```

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -172,38 +179,48 @@ def get_profile(
...
         except jsonschema.exceptions.SchemaError as exc:
+            logger.error(f"The provided schema is invalid, {exc}")
             raise jsonschema.exceptions.SchemaError(f"The provided schema is invalid, {exc}")
         except jsonschema.exceptions.UndefinedTypeCheck as exc:
+            logger.error(f"A type checker was asked to check a type it did not have registered, {exc}")
             raise jsonschema.exceptions.UndefinedTypeCheck(
                 f"A type checker was asked to check a type it did not have registered, {exc}"
             )
         except jsonschema.exceptions.UnknownType as exc:
+            logger.error(f"Unknown type is found in schema_json, exc")
```
traeok, line 203: I see that the string here is defined as an f-string - I think there are missing curly braces here:
```suggestion
            logger.error(f"Unknown type is found in schema_json, {exc}")
```

traeok (commented):
Looks good so far, thanks @pem70! I left a couple comments regarding error messaging

src/core/zowe/core_for_zowe_sdk/profile_manager.py
```diff
@@ -59,6 +60,7 @@ def __init__(self, appname: str = "zowe", show_warnings: bool = True):
         self.project_user_config = ConfigFile(type=USER_CONFIG, name=appname)
 
         self.__logger = logging.getLogger(__name__)
+        Log.registerLogger(__name__)
```
t1m0thyj, line 63: Thanks for adding the `registerLogger` function 👍  Sorry I wasn't very clear in my previous comment, but would it make sense to have the `registerLogger` function return the same logger returned by `getLogger`? That would allow us to condense these 2 lines into just 1:
```suggestion
        self.__logger = Log.registerLogger(__name__)
```

t1m0thyj (commented):
LGTM, left one more comment that is optional, and will approve once Trae's comments have been addressed 😋

pem70 (commented):
Thanks for all the comments!

t1m0thyj (approved):
LGTM, thanks @pem70!

traeok (approved):
LGTM, thanks for the logger implementation @pem70! 😁

PR #287, opened 2024-06-14
Refactor function definitions

pem70:
**What It Does**
Refactor function `create` definitions in Datasets class to avoid dictionaries as parameters.

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
@@ -380,28 +380,6 @@ class Datasets(SdkApi):
                 if options.primary is None or options.lrecl is None:
                     self.logger.error("If 'like' is not specified, you must specify 'primary' or 'lrecl'.")
                     raise ValueError("If 'like' is not specified, you must specify 'primary' or 'lrecl'.")
-
-                # if options.dsorg not in ("PO", "PS"):
-                #     self.logger.error(f"{options.dsorg} is not 'PO' or 'PS'.")
-                #     raise KeyError
-
-                # if options.alcunit is None:
-                #     options.alcunit = "TRK"
-                # elif options.alcunit not in ("CYL", "TRK"):
-                #     self.logger.error(f"{options.alcunit} is not 'CYL' or 'TRK'.")
-                #     raise KeyError
-
-                # if options.primary is not None:
-                #     if options.primary > 16777215:
-                #         self.logger.error("Specified value exceeds limit.")
-                #         raise ValueError
-                #     else:
-                #         if options.secondary is None:
-                #             options.secondary = int(options.primary / 10)
-                #         if options.secondary > 16777215:
-                #             self.logger.error("Specified value exceeds limit.")
-                #             raise ValueError
-
                 if options.dirblk is not None:
                     if options.dsorg == "PS":
                         if options.dirblk != 0:
@@ -412,20 +390,9 @@ class Datasets(SdkApi):
                             self.logger.error("Can't allocate empty directory blocks.")
                             raise ValueError
 
-                # if options.recfm is None:
-                #     options.recfm = "F"
-                # else:
-                #     if options.recfm not in ("F", "FB", "V", "VB", "U", "FBA", "FBM", "VBA", "VBM"):
-                #         self.logger.error("Invalid record format.")
-                #         raise KeyError
-
-                # if options.blksize is None and options.lrecl is not None:
-                #     options.blksize = options.lrecl
-
         custom_args = self._create_custom_request_arguments()
         custom_args["url"] = "{}ds/{}".format(self.request_endpoint, self._encode_uri_component(dataset_name))
         custom_args["json"] = options.to_dict() if options else {}
-        print(custom_args["json"])
         response_json = self.request_handler.perform_request("POST", custom_args, expected_code=[201])
         return response_json
 
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
@@ -220,15 +220,6 @@ class DatasetOption:
     def to_dict(self) -> dict:
         return {key.replace("_DatasetOption__", ""): value for key, value in self.__dict__.items() if value is not None}
 
-    def check_dirblock(self) -> bool:
-
-        if self.dsorg == "PS":
-            if self.dirblk != 0:
-                raise ValueError
-        elif self.dsorg == "PO":
-            if self.dirblk == 0:
-                raise ValueError
-
 
 class Datasets(SdkApi):
     """
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
@@ -44,14 +44,14 @@ class DatasetOption:
         self.__like = like
         self.__volser = volser
         self.__unit = unit
-        self.__dsorg = dsorg
-        self.__alcunit = "TRK" if (like == None and alcunit == None) else alcunit
+        self.dsorg = dsorg
+        self.alcunit = alcunit
         self.primary = primary
         self.secondary = secondary
         self.__dirblk = dirblk
         self.__avgblk = avgblk
-        self.__recfm = recfm
-        self.__blksize = blksize
+        self.recfm = recfm
+        self.blksize = blksize
         self.__lrecl = lrecl
         self.__storclass = storclass
         self.__mgntclass = mgntclass
@@ -81,7 +81,10 @@ class DatasetOption:
     @dsorg.setter
     def dsorg(self, dsorg: Optional[str]):
         if dsorg not in ("PO", "PS"):
-            raise KeyError
+            if self.like is not None and dsorg is None:
+                pass
+            else:
+                raise ValueError("")
         self.__dsorg = dsorg
 
     @property
@@ -90,9 +93,15 @@ class DatasetOption:
 
     @alcunit.setter
     def alcunit(self, alcunit: Optional[str]):
-        if alcunit not in ("CYL", "TRK"):
+        if alcunit is None:
+            if self.like is not None:
+                self.alcunit = None
+            else:
+                self.__alcunit = "TRK"
+        elif alcunit not in ("CYL", "TRK"):
             raise KeyError
-        self.__alcunit = alcunit
+        else:
+            self.__alcunit = alcunit
 
     @property
     def primary(self) -> Optional[int]:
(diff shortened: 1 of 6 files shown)


src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+        self.secondary = secondary
+        self.__dirblk = dirblk
+        self.__avgblk = avgblk
+        self.recfm = recfm
+        self.blksize = blksize
+        self.__lrecl = lrecl
+        self.__storclass = storclass
+        self.__mgntclass = mgntclass
+        self.__dataclass = dataclass
+        self.__dsntype = dsntype
```
t1m0thyj, line 59: Should we define all the properties without using double underscores to ensure they get validated (`self.volser = volser`, `self.unit = unit`, etc)? With the exception of `like`, I believe we discussed it's a special property and we wouldn't allow users to set it outside of the constructor.

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -152,9 +355,9 @@ def copy_dataset_or_member(
         custom_args["url"] = "{}ds/{}".format(self.request_endpoint, self._encode_uri_component(path_to_member))
         response_json = self.request_handler.perform_request("PUT", custom_args, expected_code=[200])
         return response_json
-    
-    def create(self, dataset_name, options={}):
-        """
+
+    def create(self, dataset_name, options: Optional[DatasetOption] = None):
+        """{"unit":}
```
t1m0thyj, line 360: Should this be removed?

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -163,83 +366,24 @@ def create(self, dataset_name, options={}):
...
-                    else:
-                        if options[opt] not in ("CYL", "TRK"):
-                            self.logger.error(f"{opt} is not 'CYL' or 'TRK'.")
-                            raise KeyError
-
-                elif opt == "primary":
-                    if options.get(opt) is not None:
-                        if options["primary"] > 16777215:
-                            self.logger.error("Specified value exceeds limit.")
+        if options:
```
t1m0thyj, line 369: If the `options` object is empty our validation should fail. I think we could define `options = options or {}` at the top of this function, to ensure that the object is not None, and then this `if` condition could be removed.

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+    @property
+    def secondary(self) -> Optional[int]:
+        return self.__secondary
+
+    @secondary.setter
+    def secondary(self, secondary: Optional[int]):
+        if self.primary is not None:
+            secondary = secondary if secondary is not None else int(self.primary / 10)
+            if secondary > 16777215:
+                raise ValueError
```
t1m0thyj, line 126: Can we add a unit test to validate that this error gets thrown?

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+    @dsntype.setter
+    def dsntype(self, dsntype: Optional[str]):
+        self.__dsntype = dsntype
+
+    @property
+    def like(self) -> Optional[str]:
+        return self.__like
+
+    def to_dict(self) -> dict:
+        return {key.replace("_DatasetOption__", ""): value for key, value in self.__dict__.items() if value is not None}
```
t1m0thyj, line 221: Curious why we need to replace `_DatasetOption__` here?

t1m0thyj (commented):
Looks pretty good, thanks @pem70! Left a few comments

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
pem70: Without raplacing "_DatasetOption__", the response json would be like: {"_DatasetOption__like": "None"}, potentially breaking some tests or documented format.

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
 from zowe.zos_files_for_zowe_sdk.constants import FileType, zos_file_constants
 
 _ZOWE_FILES_DEFAULT_ENCODING = zos_file_constants["ZoweFilesDefaultEncoding"]
 
+
+class DatasetOption:
+
+    def __init__(
+        self,
+        like: str = None,
```
adam-wolfe, line 27: @t1m0thyj does it make sense to have `like` as a required parameter? It seems to me like there would be times when a user would not want to base a new data set on an existing data set.

t1m0thyj: I don't think `like` is required here since there is a default value of `None`. However I agree we should probably change the type annotation to be consistent and say `Optional[str]`

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+    def dsorg(self) -> Optional[str]:
+        return self.__dsorg
+
+    @dsorg.setter
+    def dsorg(self, dsorg: Optional[str]):
+        if dsorg not in ("PO", "PS"):
+            if self.like is not None and dsorg is None:
+                pass
+            else:
+                raise ValueError("")
```
traeok: Do you think it would be helpful to add an error message when raising this `ValueError`? Not required, but it would give the user more context on why the `dsorg` assignment resulted in a failure.

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+
+    @alcunit.setter
+    def alcunit(self, alcunit: Optional[str]):
+        if alcunit is None:
+            if self.like is not None:
+                self.alcunit = None
+            else:
+                self.__alcunit = "TRK"
+        elif alcunit not in ("CYL", "TRK"):
+            raise KeyError
```
traeok, line 99: Should we show the supported formats for `alcunit` in the error message?
It's a similar request to the comment above; it's not a required change, but listing the supported types could add clarity for the error when displaying it to the user.

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+
+    @property
+    def primary(self) -> Optional[int]:
+        return self.__primary
+
+    @primary.setter
+    def primary(self, primary: Optional[int]):
+        if primary is not None:
+            if primary > 16777215:
+                raise ValueError
```
traeok, line 111: I have a similar suggestion here:

 ```suggestion
            if primary > 16777215:
                raise ValueError("Given primary space allocation exceeds track limit of 16,777,215")
```

File: src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py (hunk 4 above)
Comment (traeok) [inline src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py:123]:
```suggestion
            if secondary > 16777215:
                raise ValueError("Given secondary space allocation exceeds track limit of 16,777,215")
```

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,216 @@
...
+
+    @recfm.setter
+    def recfm(self, recfm: Optional[str]):
+        if recfm is None:
+            if self.like is not None:
+                self.__recfm = None
+            else:
+                self.__recfm = "F"
+        elif recfm not in ("F", "FB", "V", "VB", "U", "FBA", "FBM", "VBA", "VBM"):
+            raise KeyError
```
traeok, line 154: Would it be helpful to list the supported record formats in the error output? This one could go either way, I'm not opposed to omitting the supported formats from the error as there are quite a few options. But, I figured I should mention it as I had left similar comments above.

traeok (commented):
Changes LGTM, thanks @pem70 for the refactor! I left a couple comments regarding validation and error messaging.

traeok (approved):

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,215 @@
...
+
+    @property
+    def alcunit(self) -> Optional[str]:
+        return self.__alcunit
+
+    @alcunit.setter
+    def alcunit(self, alcunit: Optional[str]):
+        if alcunit is None:
+            if self.like is not None:
+                self.alcunit = None
```
adam-wolfe, line 95: Should this be self.__alcunit? I run into an issue with here when I only specify a `like` data set.

adam-wolfe: Looks good, thanks

src/zos_files/zowe/zos_files_for_zowe_sdk/datasets.py
```diff
@@ -11,13 +11,215 @@
...
+        avgblk: Optional[int] = None,
+        recfm: Optional[str] = None,
+        blksize: Optional[int] = None,
+        lrecl: Optional[int] = None,
+        storclass: Optional[str] = None,
+        mgntclass: Optional[str] = None,
+        dataclass: Optional[str] = None,
+        dsntype: Optional[str] = None,
+    ) -> None:
+        self.__like = like
```
adam-wolfe, line 44: Should we add a setter for `like` and change this to `self.like = like`?

t1m0thyj: Some of the properties like `alcunit` have different validation depending on whether `like` is set or not.

If we allow the `like` property to be changed, then we need to revalidate these other properties. For now it seemed simpler to only allow setting `like` in the constructor.

adam-wolfe: Ok that makes sense.

adam-wolfe (approved):
LGTM, thanks @pem70

zFernand0 (approved):
Hey @pem70,
Thanks for the nice refactoring! 😋 

I found a few instances where the behavior was somewhat inconsistent with other clients, but it should be resolved in a separate PR 😋 

LGTM! 😋

PR #294, opened 2024-06-24
Updated tests to use context managers

### LINKED ISSUE #288: Update tests to use context managers
Follow up to #145 - now that the SDK APIs are context managers, we should update our tests:
* [ ] All integration tests should use `with` statement when calling SDK APIs
* [ ] [core] Add unit test for SDK API that tests `requests.Session.close` is called after context manager exits

aadityasinha-dotcom:
**What It Does**

Resolves #288 

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/tests/integration/test_zos_console.py
+++ b/tests/integration/test_zos_console.py
@@ -11,7 +11,8 @@ class TestConsoleIntegration(unittest.TestCase):
     def setUp(self):
         """Setup fixtures for Console class."""
         test_profile = ProfileManager(show_warnings=False).load(profile_type="zosmf")
-        self.console = Console(test_profile)
+        with Console(test_profile) as console:
+            self.console = console
 
     def test_console_command_time_should_return_time(self):
         """Test the execution of the time command should return the current time"""
--- a/tests/integration/test_zos_console.py
+++ b/tests/integration/test_zos_console.py
@@ -12,7 +12,7 @@ class TestConsoleIntegration(unittest.TestCase):
         """Setup fixtures for Console class."""
         test_profile = ProfileManager(show_warnings=False).load(profile_type="zosmf")
         self.console = Console(test_profile)
-        self.addCleanup(lambda: self.console.__exit(None, None, None))
+        self.addCleanup(lambda: self.console.__exit__(None, None, None))
 
     def test_console_command_time_should_return_time(self):
         """Test the execution of the time command should return the current time"""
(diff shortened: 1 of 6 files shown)


pem70 (commented):
Looks pretty good to me!

tests/integration/test_zos_console.py
```diff
@@ -11,7 +11,8 @@ class TestConsoleIntegration(unittest.TestCase):
     def setUp(self):
         """Setup fixtures for Console class."""
         test_profile = ProfileManager(show_warnings=False).load(profile_type="zosmf")
-        self.console = Console(test_profile)
+        with Console(test_profile) as console:
+            self.console = console
```
t1m0thyj, line 15: For integration tests, it doesn't work to use a `with` statement in the `setUp` block because the `request.Session` object gets deleted immediately before tests can run.

I think the following would work instead:
```suggestion
        self.console = Console(test_profile)
        self.addCleanup(lambda: self.console.__exit__(None, None, None))
```

See https://stackoverflow.com/questions/8416208/in-python-is-there-a-good-idiom-for-using-context-managers-in-setup-teardown 😋

t1m0thyj (changes requested):

tests/integration/test_zos_console.py
aadityasinha-dotcom: So, the `requests.Session` object gets deleted in each integeration test

t1m0thyj: That's right, we want to delete the `requests.Session` object for each integration test, but need to wait until after it runs which can be accomplished by `addCleanup`.

tests/integration/test_zos_console.py
```diff
@@ -12,6 +12,7 @@ def setUp(self):
         """Setup fixtures for Console class."""
         test_profile = ProfileManager(show_warnings=False).load(profile_type="zosmf")
         self.console = Console(test_profile)
+        self.addCleanup(lambda: self.console.__exit(None, None, None))
```
zFernand0, line 15:
```suggestion
        self.addCleanup(lambda: self.console.__exit__(None, None, None))
```

zFernand0 (changes requested):
changes LGTM! 😋 
just one minor comment

zFernand0 (approved):

Issue #83, opened 2022-06-15
Standardize method names


t1m0thyj:
Before the Python SDK reaches the Long Term Support stage, we should evaluate all method names and make any necessary breaking changes.

t1m0thyj:
We should change `dataset` to `data set` to be consistent, and review other method names.

Also, we should consider making properties on classes private, especially for `SdkApi`.

PR #300, opened 2024-06-28
Standardize function names

pem70:
**What It Does**
Standardizes function names that `datasets` to be `data_sets`, sets unnecessary public variables to private. Addresses [#83]

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/sdk_api.py
+++ b/src/core/zowe/core_for_zowe_sdk/sdk_api.py
@@ -11,12 +11,11 @@ Copyright Contributors to the Zowe Project.
 """
 
 import urllib
-from .logger import Log
 
 from . import session_constants
+from .logger import Log
 from .request_handler import RequestHandler
 from .session import ISession, Session
-from .logger import Log
 
 
 class SdkApi:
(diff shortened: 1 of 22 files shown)


src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -11,51 +11,49 @@
...
 
-        self.request_endpoint = session.host_url + self.default_service_url
+        self._request_endpoint = session.host_url + self._default_service_url
 
-        self.request_arguments = {
-            "url": self.request_endpoint,
-            "headers": self.default_headers,
+        self._request_arguments = {
+            "url": self._request_endpoint,
+            "headers": self._default_headers,
```
traeok, line 42: I noticed that a couple of these variables are prefixed with a single underscore as opposed to double-underscore. I don't have a preference for one style over the other, but am curious to know if we want to make the style consistent for all private variables?

traeok (approved):
LGTM, thanks @pem70 for standardizing these 🙂

src/core/zowe/core_for_zowe_sdk/sdk_api.py
pem70: The variables with single underscore indicate that we should avoid using them outside, but since some other files use them anyhow, I put them in a less strict private mode so that those files can refer to them. On the other hand, the double underscore private variables are not used outside of the original file, hence I can put them as strict private.

t1m0thyj (approved):
LGTM, thanks @pem70!

Issue #296, opened 2024-06-25
[core] Session should support AUTH_TYPE_CERT


zFernand0:
Sessions should support certificates to be specified.

See the Zowe Node.js SDK: https://github.com/zowe/zowe-cli/blob/ad18d098fc2b2586d18aeb8a55ef82c69dffdcba/packages/imperative/src/rest/src/session/SessConstants.ts#L58

Issue #291, opened 2024-06-19
[core] Session should support AUTH_TYPE_NONE


t1m0thyj:
Sessions should not require user/password/tokenType/tokenValue to be defined if auth type is none.

See the Zowe Node.js SDK: https://github.com/zowe/zowe-cli/blob/ad18d098fc2b2586d18aeb8a55ef82c69dffdcba/packages/imperative/src/rest/src/session/AbstractSession.ts#L55

PR #305, opened 2024-07-05
Support AUTH_TYPE_CERT_PEM and AUTH_TYPE_NONE

pem70:
**What It Does**
Supports for `AUTH_TYPE_CERT_PEM` and `AUTH_TYPE_NONE` in `session` addressed in [#291] and [#296]

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/session.py
+++ b/src/core/zowe/core_for_zowe_sdk/session.py
@@ -82,7 +82,7 @@ class Session:
         self.session.basePath = props.get("basePath")
         self.session.port = props.get("port", self.session.port)
         self.session.protocol = props.get("protocol", self.session.protocol)
-        self.session.rejectUnauthorized = False if props.get("rejectUnauthorized") is "False" else True
+        self.session.rejectUnauthorized = False if props.get("rejectUnauthorized") == "False" else True
 
     def load(self) -> ISession:
         return self.session
--- a/src/core/zowe/core_for_zowe_sdk/session.py
+++ b/src/core/zowe/core_for_zowe_sdk/session.py
@@ -34,7 +34,6 @@ class ISession:
     tokenType: Optional[str] = None
     tokenValue: Optional[str] = None
     cert: Optional[str] = None
-    certKey: Optional[str] = None
 
 
 class Session:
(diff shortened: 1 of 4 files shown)


src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -64,9 +66,14 @@ def __init__(self, props: dict) -> None:
         elif props.get("tokenValue") is not None:
             self.session.tokenValue = props.get("tokenValue")
             self.session.type = session_constants.AUTH_TYPE_BEARER
+        elif props.get("cert") is not None and props.get("certKey") is not None:
+            self.session.cert = props.get("cert")
+            self.session.certKey = props.get("certKey")
```
zFernand0, line 71: I believe the properties we would want to use are `certFile` and `certKeyFile`, but only here.
I think it's fine to keep it as `cert` and `certKey` in the `ISession` object 😋

zFernand0 (changes requested):
TL;DR: Changes LGTM! 😋 

---

After testing the AUTH_TYPE_NONE, I'm wondering if we should update the sample SDK. I was able to use some of that code to test this functionality and it works! 🥳 
```python
from zowe.core_for_zowe_sdk.sdk_api import SdkApi

sample = SdkApi({"host": "jsonplaceholder.typicode.com"}, "/")
custom_args = sample._create_custom_request_arguments()
custom_args["json"] = {"title": "foo", "body": "bar", "userId": 10}
custom_args["url"] = f"{sample._request_endpoint}/posts"
response_json = sample.request_handler.perform_request("POST", custom_args, expected_code=[201])
print(response_json)
```

---

Tested with a local instance of the APIML (using [this instructions](https://github.com/zowe/zowe-explorer-vscode/pull/2802#pullrequestreview-2044034545)) and...

I can't seems to be able to use certificates as a valid authentication mechanism.
😢 

While trying to resolve this, I found a few other places where we may need to make changes in order to pass certs properly to the `requests.Session()` object.
And keep in mind that we would also want to make sure not to validate certs if rejectUnauthorized is False. 😋

src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -64,15 +65,24 @@ def __init__(self, props: dict) -> None:
...
+            self.session.type = session_constants.AUTH_TYPE_NONE
+            self.__logger.info("Authentication method not supplied")
+            # raise Exception("An authentication method must be supplied")
 
         # set additional parameters
         self.session.basePath = props.get("basePath")
         self.session.port = props.get("port", self.session.port)
         self.session.protocol = props.get("protocol", self.session.protocol)
-        self.session.rejectUnauthorized = props.get("rejectUnauthorized", self.session.rejectUnauthorized)
+        self.session.rejectUnauthorized = False if props.get("rejectUnauthorized") == "False" else True
```
zFernand0, line 85: I believe the comparison needs to be with a boolean and not with a string 😋 
```suggestion
        self.session.rejectUnauthorized = False if props.get("rejectUnauthorized") == False else True
```

zFernand0 (changes requested):
Left a small comment, but otherwise, the functionality LGTM! 😋

zFernand0 (approved):
LGTM! 😋

Issue #293, opened 2024-06-19
CI: Enforce that all Python files have license header


t1m0thyj:
To conform to Zowe licensing requirements, we should require that all Python files we distribute contain the EPL 2.0 header 🙂

PR #318, opened 2024-07-18
Add license header

pem70:
**What It Does**
Adds license header for files and updates workflow with auto check and addition for license headers. [#293]

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -38,6 +38,8 @@ jobs:
           pylint src --errors-only --disable=E0401,E0611 --ignore=build
           # check for lint
           pylint ./src --disable=all --enable=C0103,C0301 --ignore=build --max-line-length=127
+      - name: Check license headers
+        run: python scripts/license_header.py src
       - name: Test with pytest
         run: |
           coverage run -m pytest ./tests/unit
--- a/.github/workflows/sdk-build.yml
+++ b/.github/workflows/sdk-build.yml
@@ -33,23 +33,11 @@ jobs:
       - name: Lint with pydoclint
         run: pydoclint --exclude='.*/build/.*' src
       - name: Lint with pylint
-<<<<<<< HEAD
-        run: pylint src --disable=all --enable=C0103 --ignore=build
-      - name: Check and add license headers
-        run: python scripts/license_header.py src
-      - name: Lint with flake8
-        run: |
-          # stop the build if there are Python syntax errors or undefined names
-          flake8 ./src --count --select=E9,F63,F7,F82 --show-source --statistics
-          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-          flake8 ./src --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
-=======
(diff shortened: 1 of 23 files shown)


src/zosmf/zowe/zosmf_for_zowe_sdk/zosmf.py
traeok, line 1: There are 2 license headers at the top of this file.

.github/workflows/sdk-build.yml
```diff
@@ -34,6 +34,8 @@ jobs:
         run: pydoclint --exclude='.*/build/.*' src
       - name: Lint with pylint
         run: pylint src --disable=all --enable=C0103 --ignore=build
+      - name: Check and add license headers
+        run: python scripts/license_header.py src
```
traeok, line 38: I like the addition of a Python script for adding license headers 😋 
However, I'm wondering if this could be optimized a bit more for CI?

* Rather than making edits to the files in CI, a check for whether the script is being run in CI could save execution time. If in a CI environment, we can skip making the edits since those changes won't be pushed up anyway. As for some ideas of detecting CI, we could check for the environment variable `CI`, or the script could be updated to support a `dry-run` option (that does not make any changes).
* If in CI, the print statement "Adding license header to: " could be refactored so that it just prints the individual file paths that are missing a license header.

traeok (changes requested):
Thanks for addressing the missing licenses and for adding the script @pem70!
I left a couple comments; one is regarding the script and the other is about a duplicate license header.

src/secrets/zowe/secrets_for_zowe_sdk/__init__.py
```diff
@@ -1,3 +1,14 @@
...
+
+https://www.eclipse.org/legal/epl-v20.html
+
+SPDX-License-Identifier: EPL-2.0
+
+Copyright Contributors to the Zowe Project.
+"""
 """
 Zowe Python SDK - Client Secrets package
 """
```
t1m0thyj, line 14: Can this comment block be removed?

src/zosmf/zowe/zosmf_for_zowe_sdk/__init__.py
```diff
@@ -1,3 +1,14 @@
...
+
+https://www.eclipse.org/legal/epl-v20.html
+
+SPDX-License-Identifier: EPL-2.0
+
+Copyright Contributors to the Zowe Project.
+"""
 """
 Zowe Python SDK - z/OSMF package
 """
```
t1m0thyj, line 14: Can this comment block be removed?

traeok (approved):
LGTM, thanks @pem70!

It might be helpful to print the files even when the script isn't in "write mode," so that you can see which files have failed the check in the CI output. That said, not requesting changes for this - just a suggestion 😋

t1m0thyj (approved):
LGTM, thanks @pem70!

zFernand0 (approved):

Issue #89, opened 2022-06-17
Investigate strong typing for REST API payloads and responses


t1m0thyj:
Perhaps we can use some new Python features like type hints (in 3.5) and dataclasses (in 3.7).

Some examples of where this would be nice:
* The `options` dictionary for `create_data_set` in the zos-files SDK
* The response object returned by `submit_job` in the zos-jobs SDK

deadaf:
Hello @t1m0thyj 

This would certainly improve the codebase's readability and working with the packages easier.  I'm interested in working on this and adding type hints wherever applicable.

Before I proceed, I wanted to check if there are any specific guidelines or conventions for adding type hints in the project. Additionally, I would appreciate any guidance or considerations you might have regarding the use of type hints and dataclasses.

Please let me know your thoughts on this, and if there's anything specific I should keep in mind while working on this issue. Thank you!

t1m0thyj:
For return types, we want simple dataclasses that support both dot notation (e.g. `obj.property`) and bracket notation (e.g. `obj["property"]`) to combine the features of namespaces and dictionaries with type hints.

Here is an example from ChatGPT:
```python
from dataclasses import dataclass, field
from typing import Any


@dataclass
class ExtendedDataClass:
    attribute1: str = field(default="")
    attribute2: int = field(default=0)

    def __getitem__(self, key: str) -> Any:
        return self.__dict__[key]

    def __setitem__(self, key: str, value: Any) -> None:
        self.__dict__[key] = value


# Example usage
obj = ExtendedDataClass(attribute1="value1", attribute2=42)

# Accessing attributes
print(obj.attribute1)  # Using dot notation
print(obj["attribute1"])  # Using bracket notation

# Setting attributes
obj.attribute1 = "new_value1"  # Using dot notation
obj["attribute2"] = 84  # Using bracket notation

print(obj.attribute1)
print(obj.attribute2)
```

PR #322, opened 2024-08-07
Update REST API response type

pem70:
**What It Does**
Addresses [#89]

**How to Test**

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 ### Enhancements
 
-- *Breaking*: Update custom class for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
+- *Breaking*: Update method return types to use custom classes for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 - Added checks and auto addition for license headers on workflow. [#293] (https://github.com/zowe/zowe-client-python-sdk/issues/293)
 
+- *Breaking*: Update custom class for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
+
 ### Bug Fixes
 
 ## `1.0.0-dev18`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to the Zowe Client Python SDK will be documented in this file.
 
+## Recent Changes
+
+### Enhancements
+
+- *Breaking*: Update custom class for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
+
+### Bug Fixes
+
+
 ## `1.0.0-dev19`
 
 ### Enhancements
@@ -13,8 +22,6 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 - *Breaking*: Revised function names in `Logger` class and `files` class into snake_case. Enabled pylint rule to enforce function case. [#315] (https://github.com/zowe/zowe-client-python-sdk/issues/315)
 
 - Added checks and auto addition for license headers on workflow. [#293] (https://github.com/zowe/zowe-client-python-sdk/issues/293)
-
-- *Breaking*: Update custom class for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
 
 ### Bug Fixes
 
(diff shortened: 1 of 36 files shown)


zFernand0 (commented):
LGTM so far! 😋 
Noticed a few untouched lines, mostly around the getitem and setitem, but I assume that's just because of how most unit tests might be referencing properties using dot notation as oposed to bracket notation (or vice-versa) 😋

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 All notable changes to the Zowe Client Python SDK will be documented in this file.
 
+## Recent Changes
+
+### Enhancements
+
+- *Breaking*: Update custom class for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
```
t1m0thyj, line 9: Should we mention something about method return types here?
```suggestion
- *Breaking*: Update method return types to use custom classes for REST API responses [#89] (https://github.com/zowe/zowe-client-python-sdk/issues/89)
```

src/_version.py
t1m0thyj, line 1: Since this file is automatically generated by [our release workflow](https://github.com/zowe/zowe-client-python-sdk/blob/9c42f3b57372ac92e34955558a619ee31f34bfc3/.github/workflows/sdk-release.yaml#L46-L47), we may want to exclude it from the script that adds license headers.

src/zos_console/zowe/zos_console_for_zowe_sdk/console.py
```diff
@@ -40,17 +42,17 @@ def issue_command(self, command: str, console: Optional[str] = None) -> dict:
...
         custom_args = self._create_custom_request_arguments()
         custom_args["url"] = self._request_endpoint.replace("defcn", console or "defcn")
         request_body = {"cmd": command}
         custom_args["json"] = request_body
         response_json = self.request_handler.perform_request("PUT", custom_args)
-        return response_json
+        return IssueCommandResponse(response_json)
 
-    def get_response(self, response_key: str, console: Optional[str] = None) -> dict:
+    def get_response(self, response_key: str, console: Optional[str] = None) -> GetReponseResponse:
```
t1m0thyj, line 55: `GetReponseResponse` -> `GetResponseResponse`

If this sounds too redundant, we could change it to a more generic name like `TsoDataResponse` 😋

src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, response: dict) -> None:
+        for k, value in response.items():
+            key = k.replace("-", "_")
+            super().__setattr__(key, value)
+
+    def __getitem__(self, key: str) -> str:
+        return self.__dict__[key.replace("-", "_")]
+
+    def __setitem__(self, key: str, value: str) -> None:
+        self.__dict__[key.replace("-", "_")] = value
```
t1m0thyj, line 32: These seem like common functions that are shared across all of the response dataclasses. If so, does it make sense to move them into a base class in the Core SDK that all the other classes can inherit?
```python
class BaseResponse:
    def __init__(self, response: dict) -> None:
        for k, value in response.items():
            key = k.replace("-", "_")
            super().__setattr__(key, value)

    def __getitem__(self, key: str) -> str:
        return self.__dict__[key.replace("-", "_")]

    def __setitem__(self, key: str, value: str) -> None:
        self.__dict__[key.replace("-", "_")] = value
```

src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@dataclass
+class GetReponseResponse:
```
t1m0thyj, line 36: Same comment as above about the name of this class

t1m0thyj (changes requested):
Looks pretty good, thanks @pem70 for adding all these return types 😍

src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
pem70: Only three classes use the exact same common functions. Other classes need to adjust this code based on their attributes. Although I like this idea, I hardly think it would be that helpful to create a separate base class.

traeok (approved):
Changes LGTM, thanks Peizhao for adding these types!

tests/integration/test_zos_files.py
```diff
@@ -46,24 +50,24 @@ def test_list_dsn_should_return_a_list_of_datasets(self):
...
             self.assertGreater(len(command_output["items"]), 0)
 
             # Assert that the first item in the list has 'dsname' defined
             first_item = command_output["items"][0]
-            self.assertIn("dsname", first_item)
+            self.assertTrue(hasattr(first_item, "dsname"))
 
             # Assert that the first item in the list has the expected attributes defined
-            attributes = first_item.keys()
+            attributes = dir(first_item)
```
t1m0thyj, line 63: It might be nice if we implemented standard dictionary methods like `.keys`, `.values` etc on our response data classes, but this could be done as a future enhancement 😋

t1m0thyj (approved):

awharn (approved):

PR #363, opened 2025-05-20
feat: added pytest fixture to create and use its own temporary directory structure

### LINKED ISSUE #361: [tests]: Make integration tests not dependent on the dev's environment
**Is your feature or enhancement request related to a problem or limitation? Please describe**

It's be good to configure integration tests in such a way that they are not impacted by the developer's environment

**Describe your enhancement idea**

Create an isolated Project + Global directory combination for testing purposes

**Describe alternatives you've considered**

Right now, devs may be forces to alter their environment in order to run integration tests

**Provide any additional context**

N/A

rohansen856:
**What It Does**

Fixes #361 

## Summary
As discussed in issue #361  I have added two files inside the tests directory which contains pytest fixtures to create and use a `tmp_dir` for the integration tests to run. This is the first part of incrementally implementing the directory isolation functionality for the integration tests.

@zFernand0  If this fix is alright, For the next steps I would like to Implement `monkeypatch` to temporarily change the Environment variables inside the `tmp_dir` as to ascertain that the dev's local environment is not clashing with the tests.

**How to Test**

- run the tests using
```sh
 pytest ./tests/integration 
```

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**


tests/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+def isolated_test_env(tmp_path):
+    project_dir = tmp_path / "project"
+    global_dir = tmp_path / "global"
+    project_dir.mkdir()
+    global_dir.mkdir()
+
+    return {
+        "project_dir": project_dir,
+        "global_dir": global_dir
+    }
```
zFernand0, line 14: Thanks for keeping the globa vs. project configs here...
I believe that for the purpose of integration testing we could simplify this to just use the 
`tmp_path / zowe.config.json`
```suggestion
    config_path = Path(__file__).resolve().parent.parent.parent / "zowe.config.json"
    config_file = tmp_path / "zowe.config.json"
    config_file.write_text(config_path.read_text())
    return tmp_path
```

zFernand0 (commented):
This is a great first step! 🥳 

I'd love to see this being leveraged by other tests. In order to make this easier to maintain... we could move the creating of the zowe.config.json file in the tmp_path to the isolated_test_env function. Then we could create a class in `conftest.py` that other test classes can leverage...
Here is an example...
```python
class TestIsolatedEnv(unittest.TestCase):
    """Base class for isolated test environments."""
    @pytest.fixture(autouse=True)
    def _setup_test_env(self, isolated_test_env):
        self.isolated_test_env = isolated_test_env
```

And then other tests can easily make use of the isolated_test_env by inheriting from it
```python
class TestConsoleIntegration(TestIsolatedEnv):
```

Also, for easier import, we may need to move conftest.py inside the integration folder (but that may cause other problems 😋 
So we might be forced into creating a `zowe.test_utils` package for this class 😓 

---

The above is just one way (of many) that we could implement this.

zFernand0:
Hey @rohansen856,
Please note that all of my comments above are mere suggestions 😋 

I'm sure there are better ways to avoid duplicating the creation/copy of the "project" zowe.config.json file into the tmp_path 😋 

I'd love to get @t1m0thyj's thoughts on this 😋

pem70 (commented):
LGTM. Same suggestions as @zFernand0 gave.

rohansen856:
> This is a great first step! 🥳
> 
> I'd love to see this being leveraged by other tests. In order to make this easier to maintain... we could move the creating of the zowe.config.json file in the tmp_path to the isolated_test_env function. Then we could create a class in `conftest.py` that other test classes can leverage... Here is an example...
> 
> ```python
> class TestIsolatedEnv(unittest.TestCase):
>     """Base class for isolated test environments."""
>     @pytest.fixture(autouse=True)
>     def _setup_test_env(self, isolated_test_env):
>         self.isolated_test_env = isolated_test_env
> ```
> 
> And then other tests can easily make use of the isolated_test_env by inheriting from it
> 
> ```python
> class TestConsoleIntegration(TestIsolatedEnv):
> ```
> 
> Also, for easier import, we may need to move conftest.py inside the integration folder (but that may cause other problems 😋 So we might be forced into creating a `zowe.test_utils` package for this class 😓
> 
> The above is just one way (of many) that we could implement this.

Thanks @zFernand0 😁... I tested the functions with the `conftest.py` file inside `/integration` folder and it the results were same as before (the fixtures affect them the same way in booth cases), so I think we can definitely move the file as you suggested 🫡. 
Pardon me but i didn't get the point of this: `class TestConsoleIntegration(TestIsolatedEnv):` It would be great if you can possibly elaborate a bit (we need to take verything into consideration as all the integration tests would depend on it i guess). Thanks for all the suggestions. Lets work on it to find the optimal way🫡.

tests/conftest.py
rohansen856: I think instead of `tmp_path` we need to return the `config_file` and `config_path` variables from the function. It would look something like this:
```py
from pathlib import Path

import pytest


@pytest.fixture
def isolated_test_env(tmp_path):
    config_path = Path(__file__).resolve().parent.parent.parent / "zowe.config.json"

    config_file = tmp_path / "zowe.config.json"
    config_file.write_text(config_path.read_text())

    return {
        "config_path": config_path, 
        "config_file": config_file
    }
```

And for the other file (`tests/integration/test_integration_env.py`):
```py
def test_isolated_env_does_not_touch_real_home(isolated_test_env):
    config_path = isolated_test_env["config_path"]
    config_file = isolated_test_env["config_file"]

    assert config_path.exists(), f"Missing config at {config_path}"
    assert config_file.exists()
    assert "profiles" in config_file.read_text()
```

And yes this simplifies the overall structure by isolating all the logic and assertions in different functions🫡. Please let me know if I am mistaken anywhere...

rohansen856: If this is alright then we can possibly move towards making the structure more modular by using it like this: `class TestIsolatedEnv(unittest.TestCase):` as you suggested. A step by step approach would be better so I would like to first make a commit for modifications as mentioned above then move towards the modular integration...

zFernand0:
Hey @rohansen856,
Thanks for considering the suggested implementation. 
Please let us know when you get a chance to work on those (either here or in a follow-up PR) 😋

rohansen856:
@zFernand0 I will start working on these from Wednesday at the earliest. I would like to work on this very PR as a draft (a lot more discussions might be ahead and we need to keep track of them in a single thread) and after all the things are finalized I can either request to merge this PR (if the overall structure is still clean at the end) or open up a fresh PR with all the changes in a systematic order...
Thanks!

rohansen856:
@zFernand0 first of all sorry for the late!

I have updated asll the integration test classes with the above discussed method (ex.: `TestZosmfIntegration(unittest.TestCase)` ->TestZosmfIntegration(TestIsolatedEnv)`) and the tests are seen to be passing!

tests/conftest.py
t1m0thyj: @zFernand0 is it ok to require `zowe.config.json` to be at the root level of the project, or should we allow it to reside at any folder level?

tests/integration/test_integration_env.py
```diff
@@ -0,0 +1,9 @@
+"""Test isolated environment setup for integration tests."""
+
+def test_isolated_env_does_not_touch_real_home(isolated_test_env):
```
traeok, line 3: - Does this actually verify that the environment does not touch the "real home"?
- When you mean "real home," are you referring to the user's home directory? or `ZOWE_CLI_HOME`?

tests/integration/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+
+    config_file = tmp_path / "zowe.config.json"
+    config_file.write_text(config_path.read_text())
+
+    return {
+        "config_path": config_path, 
+        "config_file": config_file
+        }
+
+class TestIsolatedEnv(unittest.TestCase):
```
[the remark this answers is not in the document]
traeok: I think we should consider this comment as we don't want this to negatively impact test collection behavior. Maybe we can pursue those xunit setup hooks.

traeok (commented):
Thanks for your work on this @rohansen856. I've left some comments - I believe we should also address the comments posted by Copilot.

zFernand0 (commented):
I really like this PR! 🙏 
Just waiting for some of the comments to be address 🙏

PR #369, opened 2025-06-19
Support responseTimeout profile property for z/OSMF operations

### LINKED ISSUE #334: Add a way to use a secure profile to connect to the remote z/OS system
The connection function only supports specific user name and password input and no way to use a secure profile name.

Provide either a new argument to the connection function or a new "profile_connect" function to connect to the remote system using a config.json profile with secure user credentials.

In the SDK documentation there does not seem to be any way to use a loaded profile to connect to the remote system.

aadityasinha-dotcom:
**What It Does**
Fixes #334 

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] tested my changes
- [ ] updated the changelog

**Additional Comments**


src/core/zowe/core_for_zowe_sdk/session.py
```diff
@@ -67,6 +68,7 @@ def __init__(self, props: dict[str, Any]) -> None:
             self.session.user = props.get("user")
             self.session.password = props.get("password")
             self.session.reject_unauthorized = bool(props.get("rejectUnauthorized"))
+            self.session.response_timeout = props.get("responseTimeout")
```
t1m0thyj, line 71: @zFernand0 @traeok Is it ok to store `response_timeout` on the `Session` class since it's specific to z/OSMF? Wondering if we should keep the properties on the `Session` class more profile/service-agnostic.

pem70 (commented):
Changes LGTM.

zFernand0 (changes requested):
Changes make sense.
I agree with Timothy's comment that we should keep the session free from z/somf specific properties

src/core/zowe/core_for_zowe_sdk/session.py
aadityasinha-dotcom: Thank you for the review, can you review the changes I tried moving it to the zosmf package :relaxed:

zFernand0 (dismissed):
LGTM! 😋 
---

I do have one small request...

Doing this only in the zosmf package will impact only the functions dependent on the zosmf...

From the IBM documentation...
https://www.ibm.com/docs/en/zos/2.5.0?topic=services-zos-data-set-file-rest-interface#izuhpinfo_api_restfiles__title__5

I believe we would want to add this same funcitonality in the `files.py`.
And perhaps also add some validation around it (mostly to warn people that the value may be modified by z/OSMF
`5 >= responseTimeout >= 600` 😋 

Image

If possible, please keep in mind that more headers may need to be introduced in the future 🙏

src/zos_files/zowe/zos_files_for_zowe_sdk/files.py
```diff
@@ -26,6 +26,8 @@
 
 _ZOWE_FILES_DEFAULT_ENCODING = zos_file_constants["ZoweFilesDefaultEncoding"]
 
+_MIN_TIMEOUT = 5
+_MAX_TIMEOUT = 600
```
t1m0thyj: Can we define these inside of `zos_file_constants`?

tests/unit/test_zosmf.py
```diff
@@ -35,3 +35,26 @@ def test_list_systems(self, mock_send_request):
 
         Zosmf(self.connection_dict).list_systems()
         mock_send_request.assert_called_once()
+
+    @mock.patch("requests.Session.send")
+    def test_response_timeout_header_added(self, mock_send_request):
```
t1m0thyj: After addressing previous comment, please move this test to `test_files.py`.

t1m0thyj (changes requested):

t1m0thyj (changes requested):
Code changes LGTM! Please ~~resolve failing unit tests~~ (I've done this 😁) and add a changelog entry 😋

src/zos_files/zowe/zos_files_for_zowe_sdk/api.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+
+    def __init__(self, profile: dict[str, Any], log: bool = True):
+        super().__init__(profile, "/zosmf/restfiles/", logger_name=__name__, log=log)
+        self._default_headers["Accept-Encoding"] = "gzip"
+        self._set_response_timeout(profile)
+
+    def _set_response_timeout(self, profile: dict[str, Any]):
+        async_threshold = profile.get("asyncThreshold") or profile.get("X-IBM-Async-Threshold")
+        if not async_threshold:
```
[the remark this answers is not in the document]
aadityasinha-dotcom: That's true, I also think `asyncThreshold` check can be removed.

src/zos_files/zowe/zos_files_for_zowe_sdk/api.py
```diff
@@ -0,0 +1,56 @@
...
+                        f"responseTimeout {resp_to_int} out of range; clamped to {clamped} "
+                        f"(allowed {_MIN_TIMEOUT}-{_MAX_TIMEOUT})"
+                    )
+                self._default_headers["X-IBM-Response-Timeout"] = str(clamped)
+            except (TypeError, ValueError):
+                if hasattr(self, "logger"):
+                    self.logger.warning("responseTimeout must be an integer between 5 and 600; header not set")
+        else:
+            if hasattr(self, "logger"):
+                self.logger.info("X-IBM-Async-Threshold present; X-IBM-Response-Timeout will be ignored by z/OSMF")
```
t1m0thyj: Please remove the `else` block - we don't need to log a warning when `responseTimeout` is missing on the profile. Everything else LGTM 😋

t1m0thyj (approved):
LGTM, thanks @aadityasinha-dotcom!

zFernand0 (approved):
LGTM! 😋

traeok (dismissed):
Hi @aadityasinha-dotcom , I noticed that the build pipelines did not succeed due to a failing unit test:

```
        prepared = mock_send_request.call_args[0][0]
>       self.assertIsNone(prepared.headers.get("X-IBM-Response-Timeout"))  # ignored when async header present [web:102]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AssertionError: '120' is not None
```

Please resolve this when you get the chance. Thanks

zFernand0 (approved):
LGTM! 😋

PR #372, opened 2025-07-09
Added zOSMF Workflows API

KUGDev:
**What It Does**

Added z/OSMF Workflows module.

**How to Test**

See integration `tests/integration/test_zosmf_workflows.py` and `zowe/zosmf_workflows/README.md` for the usage examples.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [ ] updated the changelog

**Additional Comments**

Ready to introduce any relevant changes. Notifications REST API functionality is not implemented, so to check a workflow progress you need to use built-in libraries (e.g. `times` for the thread sleep).


src/zosmf_workflows/setup.py
```diff
@@ -0,0 +1,42 @@
...
+
+https://www.eclipse.org/legal/epl-v20.html
+
+SPDX-License-Identifier: EPL-2.0
+
+Copyright Contributors to the Zowe Project.
+
+Contributors:
+    Zowe Community
+    Uladzislau Kalesnikau
```
zFernand0, line 14: I haven't seen this before from Zowe core contributors, but that's cool! 😅 

Is this something we want to start doing going forward? 🤔 

If so, we may want to adjust some license_header scripts on other repos.

Personally, I'd vote for sticking to the Git sign-offs instead of adding individuals names to each file since it can get out of hand 😅 
 
If someone is concerned about the legitimacy of git sign-offs, there is the possibility to configure GPG signing for each commit made to the repo! 🙏

src/zosmf_workflows/zowe/zosmf_workflows_for_zowe_sdk/workflows.py
```diff
@@ -0,0 +1,557 @@
...
+        comments: Optional[str] = None,
+        assign_to_owner: bool = True,
+        access_type: Literal['Public', 'Restricted', 'Private'] = "Public",
+        account_info: Optional[str] = None,
+        job_statement: Optional[str] = None,
+        delete_completed_jobs: bool = False,
+        jobs_output_directory: Optional[str] = None,
+        auto_delete_on_completion: bool = False,
+        target_systemuid: Optional[str] = None,
+        target_systempwd: Optional[str] = None
```
zFernand0, line 88: Thanks for keeping the similar defaults as the Nodejs SDK and the z/osmf REST API. 🙏 
I do wonder about these two properties in particular.

I know these are part of the request body, but curious to get your thoughts on the other endpoints that support actions on remote systems. (see [X-IBM-Target-System](https://www.ibm.com/docs/en/zos/3.1.0?topic=services-zos-data-set-file-rest-interface#izuhpinfo_api_restfiles__title__5)) 😋

zFernand0 (changes requested):
Changes LGTM so far! 😋 
---

I do have some requested changes to consider. 🙏 
Also, I may leave more comments as I continue reading through the code and testing the functionality 🙏

tests/integration/test_zosmf_workflows.py
```diff
@@ -0,0 +1,209 @@
...
+        self.assertEqual(len(workflow_def.steps), 2)
+        self.assertEqual(len(workflow_def.variables), 3)
+
+        self.created_workflow = self.workflows.create_workflow(
+            workflow_name="Zowe Test Workflow (Common)",
+            workflow_definition_file=self.test_uss_workflow,
+            system="BT41",
+            owner="gast88n",
+            variable_input_file=self.test_uss_workflow_input,
+            workflow_archive_safid="gast88n"
```
zFernand0, line 81: We should load these values from fixtures files.
I believe we might need something like the `.env.example` file where people can copy it and provide their own set of information.

zFernand0 (changes requested):

tests/integration/fixtures/workflow.xml
```diff
@@ -0,0 +1,71 @@
...
+        <setVariable name="ZOWE_TEST_VAR_3" scope="instance" substitution="true">${instance-ZOWE_TEST_VAR_1}+${instance-ZOWE_TEST_VAR_2}</setVariable>
+    </step>
+    <step name="fake-check" optional="true">
+        <title>Zowe test step 2</title>
+        <description>Fake JCL job run try</description>
+        <prereqStep name="variables"/>
+        <instructions substitution="false">Nothing special, just a fake JCL job run that will fail</instructions>
+        <weight>1</weight>
+        <autoEnable>false</autoEnable>
+        <canMarkAsFailed>false</canMarkAsFailed>
```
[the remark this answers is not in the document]
KUGDev: I agree with the `autoEnable`. I can do so for `canMarkAsFailed`, but I don't think it will have any effect:
изображение

tests/integration/test_zosmf_workflows.py
```diff
@@ -0,0 +1,209 @@
...
+                self.workflows.delete_workflow(self.created_workflow.workflowKey)
+            except:
+                pass
+            try:
+                self.workflows.delete_archived_workflow(self.created_workflow.workflowKey)
+            except:
+                pass
+
+    def test_get_definition_create_get_properties_list_start_cancel_delete(self):
+        """Get workflow definition, create a workflow, get it's properties, list workflows by the name regex, start the workflow, cancel and delete it"""
```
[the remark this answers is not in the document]
KUGDev: I'm not sure it makes sense here. Cause actually, to test the workflow, it needs to be created first => so it should be deleted after it is created. This part will be the same for each step. The rest of the steps: get workflow definition, get workflow properties, list workflows, and the separate area - archived workflows, that would require the same steps with create / delete. I can do it as separate steps, if needed, but still, I don't see the actual benefit in this action. Only to stress-test the system.

KUGDev:
Hi @zFernand0 
Thank you for the review!
I pushed some changes, for the rest - please, see my replies

README.md
```diff
@@ -149,6 +149,7 @@ Alternatively, you can use an existing Zowe CLI profile instead:
 
 Currently, the Zowe Python SDK supports the following interfaces:
 
+- z/OSMF Workflows
```
zFernand0, line 152: (nitpick) Could we move this to the bottom? 

My intention is to try to keep things in the order in which they were added. However, I don't thikn Console commands were the first thing to be added...
Perhaps we just sort them alphabetically 🙏

README.md
zFernand0, line 1: I'd love to get @anaxceron's review on this README file 🙏 

We could address suggestions in a follow up PR 🙏

src/workflows/README.md
zFernand0, line 1: Would like to get @anaxceron's review on this when possible 🙏

tests/integration/test_workflows.py
zFernand0, line 1: I understand we don't want to split up these two integration test cases...
But, could we reduce the verbosity of the tests?

For example...
We get the full HTTP responses even when tests pass. 
Could we suppress some of those by default in the tests (and only print them out if the test failed)?

Output of this test file... 
Image
Each line above has 160+ characters (for only two passing tests) 😅 

Output of other tests...
Image

src/workflows/zowe/workflows_for_zowe_sdk/response/workflows.py
zFernand0, line 1: This looks great! 
Thanks for the detailed documentation of the responses! 🙏 

I do have a few suggestions... 😋 
- Consider combining shared properties into common classes.
    - This should help reduce the size of this file, thus making it easier to maintain in the future 🙏 
- Consider omitting the description for self explanatory fields (e.g. workflowDescription)
- Should we add new-lines ourselves, or should we let the IDEs do the wrapping themselves? 
    - Meaning, we could just have long one liners instead of arbitrarily breaking up the description
    - VScCode does a good job ad breaking it down (since it just removed the new lines added by us) - pictures below



Image

Image

Image

zFernand0 (changes requested):
LGTM! 😋 
---

I do have a few small suggestions to consider! 🙏 

The reason for requesting changes is mainly the `env.yml` file and how it can be accidentally added to source by developers in the future 🙏

zFernand0:
Forgot to mention...

I do have some integration tests failing:
- test_mount_unmount_zfs_file_system
- test_get_file_content_streamed_should_return_response_content
- test_issue_command_should_return_valid_response (works when executed by itself)
- test_copy_uss_to_data_set_should_be_possible (works when executed by itself)

However, they fail in the same fashion in the main branch.
So, this could be a problem with my environment 🙏

CHANGELOG.md
zFernand0, line 1: Please move these new entries under a new `## Recent Changes` header 🙏

tests/integration/test_workflows.py
KUGDev: @zFernand0 I'm unsure about these messages, cause I don't get them myself running the tests...
The screenshot I provided - these are the only messages I have.
Most probably those are the messages from some other functions that are running differently on different systems (I think it is a `uss.upload` change causing them)
Could you provide a more detailed info on what kind of messages you are receiving?

изображение

src/workflows/zowe/workflows_for_zowe_sdk/response/workflows.py
KUGDev: @zFernand0 I will remove the unnecessary newlines.
About the generalization: I tried to do so, and I thought about doing it in the first place. But when I made some significant progress on that, I figured out that there are some differences in the basic properties, despite the fact that they have the same names...
If you look into some of them in their respective references, you can find some differences in such places as "Get workflow properties" and "Get workflow definition" (`isCallable` parameter, `steps` and `variables` parameters). Also, some of the parameters are common for one type of response classes, other - for different other type of response classes. In that situation, some of the properties could be found in `ListWorkflowsResponse`, `GetWorkflowPropertiesResponse`, and `GetWorkflowDefinitionResponse` classes, but are not in the `CreateWorkflowResponse` class (`workflowDefinitionFileMD5Value` property). If we are talking about the `workflowKey` property (the most interesting one), then it is common for the created workflows, but not for the workflow definitions themselves (that is correct, as soon as the definitions are not the created workflows). So doing such thing as providing some abstraction in this situation can cause a lot of misunderstanding in how and what to put to the base class...
The dataclass functionality provides some simplifications of working with such property classes, and if we provide some other way of working with them, it can cause an overhead. If we go with multi-inheritance (that is basically an antipattern), then it can solve some of the listed problems, but still could be an overhead...
So I decided to stick with the separate classes in here (until we have a better solution?).
First of all - these classes are the response classes that just describe the whole structure to be returned by z/OSMF REST API. Making abstractions in such situation can potentially increase the cognitive complexity of the solution.
Second of all - these classes potentially are the subject of change in the future by z/OSMF REST API team. The structure we have right now is the best way to reflect to these changes.

Anyways, let me know if there are any ideas on the structure, and maybe there is a better way of working with it:)

KUGDev: @zFernand0 also, about the self-explanatory fields.
I introduced the documentation for each of them respectively the same way it is done in the IBM docs. 
Also, VSCode provides the way to see the docs without going to the original place, where the class is defined, so it is visible like this:
изображение

So, if there is a parameter that is not mentioned in the documentation, it is omitted in the brief description, that could cause a misunderstanding of the actual purpose of the class. I think, it is better to have all the parameters described (even the self-explained ones) for the consistency...

CHANGELOG.md
t1m0thyj: The `1.0.0-dev23` version has already been released. Please add a new section in the changelog formatted like this:
```md
## Recent Changes

### Enhancements

- Enhanced abc

### Bug Fixes

- Fixed xyz
```

src/workflows/zowe/workflows_for_zowe_sdk/response/workflows.py
zFernand0: Hey @KUGDev,
I agree about potentially causing confusion as to what's needed in the base class 😅 

---

> these classes potentially are the subject of change in the future by z/OSMF REST API team

I believe z/OSMF may be in maintenance mode 😅 , so I don't really expect them to break any of these APIs.

---


> Anyways, let me know if there are any ideas on the structure, and maybe there is a better way of working with it:)

We can handle any refactoring in future PRs. 🙏

tests/integration/test_workflows.py
```diff
@@ -17,7 +17,7 @@
 
 FIXTURES_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "fixtures")
 FILES_FIXTURES_PATH = os.path.join(FIXTURES_PATH, "files.json")
-ENV_FIXTURE_PATH = os.path.join(FIXTURES_PATH, "env.yml")
+ENV_FIXTURE_PATH = os.path.join(FIXTURES_PATH, "env.example.yml")
```
zFernand0, line 20: I believe we might want to keep this as `env.yml` that way people can just create/update their env.yml file instead of having to be cautious about not committing their env.example.yml file

tests/integration/test_workflows.py
zFernand0: Hey @KUGDev 

As part of the setup... Right before the test starts...
I get two 404 errors, could be environmental or perhaps due to the system being too fast (or too slow?)

```
HTTP Request has failed with status code 404. 
 https://lpar.1:1234/zosmf/workflow/rest/1.0/workflows/a3eafb90-a46f-4273-bb26-339278ee23ab
```

`src/core/zowe/core_for_zowe_sdk/request_handler.py`
```
self.__logger.error(f"HTTP Request has failed with status code {self.__response.status_code}. \n {output_str}")
```

the tests work though 🙏

zFernand0 (changes requested):
LGTM! 😋 
---

one small change requested...

`ENV_FIXTURE_PATH = os.path.join(FIXTURES_PATH, "env.yml")`

.env.example
```diff
@@ -1,2 +1,5 @@
 # Required for automated integration testing
 ZOWE_TEST_PROFILE=''
+ZOS_WORKFLOWS_SYSTEM=SYST
+ZOS_WORKFLOWS_OWNER=MYUSER
+ZOS_WORKFLOWS_SAF_ID=MYUSER
```
zFernand0: For consistency, can we either remove the quotes above, or add quotes on the new ones?

```suggestion
ZOWE_TEST_PROFILE=zosmf
ZOS_WORKFLOWS_SYSTEM=SYST
ZOS_WORKFLOWS_OWNER=MYUSER
ZOS_WORKFLOWS_SAF_ID=MYUSER
```

PR #391, opened 2026-05-14
get_log_messages console REST API function, other enhancements

KUGDev:
**What It Does**

Added get_log_messages console REST API function, added exec-data and status query parameters to list_jobs, corrected members parsing, list_members enhancement

**How to Test**

Use `Console.get_log_messages` to get messages from SYSLOG/OPERLOG.
Use `fetch_exec_data=True` flag with `Jobs.list_jobs` to fetch `exec-data` with the list of jobs.
User `status='ACTIVE'` flag with `Jobs.list_jobs` to fetch jobs that are active at the moment.

**Review Checklist**
I certify that I have:
- [x] tested my changes
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
+++ b/src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
@@ -10,8 +10,30 @@ SPDX-License-Identifier: EPL-2.0
 Copyright Contributors to the Zowe Project.
 """
 
+import re
 from dataclasses import dataclass
+from datetime import datetime, timedelta
+from datetime import timezone as tz
 from typing import Any, Optional
+
+CAMEL_TO_SNAKE_CASE_PATTERN = re.compile(r'(?<!^)(?=[A-Z])')
+
+
+def to_snake_case(src_str: str) -> str:
+    """
+    Convert camel case string (cameCaseString) to a snake case string (snake_case_string).
+
+    Parameters
+    ----------
+    src_str : str
+        The string to convert
+
+    Returns
+    -------
+    str
+        The converted string
+    """
+    return CAMEL_TO_SNAKE_CASE_PATTERN.sub('_', src_str).lower()
 
 
 @dataclass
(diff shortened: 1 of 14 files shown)


src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
```diff
@@ -48,3 +70,139 @@ def __getitem__(self, key: str) -> Any:
...
+        Text that describes the cause of the error.
+    """
+    return_code: Optional[int] = None
+    reason_code: Optional[int] = None
+    reason: Optional[str] = None
+    
+    def __init__(self, response: dict[str, Any]) -> None:
+        for raw_key, value in response.items():
+            key = to_snake_case(raw_key)
+        super().__setattr__(key, value)
```
zFernand0: Shouldn't we be calling `setattr` inside the for loop?
```suggestion
            super().__setattr__(key, value)
```

I could be mistaken, but it seems that only the last property will get set 😅

tests/unit/files/datasets/test_create.py
```diff
@@ -42,6 +42,110 @@ def test_create_data_set_with_like(self, mock_send_request: mock.Mock):
         Files(self.test_profile).create_data_set("DSNAME123", options=option)
         self.assertEqual(mock_send_request.call_count, 2)
 
+    @mock.patch("logging.Logger.error")
+    @mock.patch("requests.Session.send")
+    def test_create_data_set_po_empty_dirblk(self, mock_send_request, mock_logger_error: mock.MagicMock):
```
zFernand0: This seems to be duplicated down in line 67

Should this one be renamed to `test_create_data_set_po_zero_dirblk`? (or something similar?)

zFernand0 (dismissed):
Changes are looking good.

I still have to do some more testing, and I may have more comments later on, but so far so good.

I do have a few small requests 🙏

src/zos_console/zowe/zos_console_for_zowe_sdk/response/console.py
KUGDev: Yes, good catch, fixed!

tests/unit/files/datasets/test_create.py
KUGDev: Looks like I mistakenly copied this one, removed in the latest commit

PR #403, opened 2026-07-28
fix job-output path handling 

ATorrise:
**What It Does**
- Fixes `get_job_output_as_files` which was writing to a folder it never created and crashing with FNF
- Builds the output paths so they match the folders the method actually makes
- Checks job values from the response before using them as folder names
- Rejects any name with .. in it (team discussion request)
- Rejects any name that is an absolute path (team discussion request)
- Makes sure the final file path stays inside the output folder

**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog
- [X] manually tested my changes
- [X] added/updated automated unit/integration tests

**Additional Comments**

### DIFF
--- a/src/zos_files/zowe/zos_files_for_zowe_sdk/uss.py
+++ b/src/zos_files/zowe/zos_files_for_zowe_sdk/uss.py
@@ -149,13 +149,13 @@ class USSFiles(BaseFilesApi):  # type: ignore
         ----------
         file_path: str
             Path of the file
-        content_type: ContentType, optional
+        content_type: ContentType
             The content type to receive ("text" or "binary", "text" by default)
-        remote_file_encoding: str, optional
+        remote_file_encoding: str
             Encoding file content originally in (to convert from; by default, it is always being converted from "IBM-1047")
-        receive_in_encoding: str, optional
+        receive_in_encoding: str
             Encoding to convert file content to (to convert to; by default, it is always being converted to "ISO8859-1")
-        as_stream: bool, optional
+        as_stream: bool
             Specifies whether the response is streamed. Default: False
 
         Returns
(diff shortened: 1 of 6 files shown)


src/zos_files/zowe/zos_files_for_zowe_sdk/uss.py
zFernand0, line 143: Hey, why are we removing the `optional` from the pydoc?
Is that because of a change in pydoclint?
We still have the defaults provided in the function signature. 🤔 

---

Yeah, just saw the commit:

Image

---

I guess none of these changes will break the usage of the functions:

Code:
Image

Image

Result:
Image

src/zos_jobs/zowe/zos_jobs_for_zowe_sdk/jobs.py
```diff
@@ -440,6 +440,32 @@ def get_spool_file_contents(self, correlator: str, id: str) -> str:
...
+        segments = relative_path.split(os.sep) if relative_path else []
+        if not segments or ".." in segments or os.path.isabs(relative_path):
+            return False
+        return True
+
+    @classmethod
+    def _reject_unsafe_path(cls, base_dir: str, target: str) -> None:
+        # Final guard that the generated path stays inside base_dir before any file is written
+        if not cls._is_sub_path(base_dir, target):
+            raise ValueError("The generated file path is outside the output directory: {}".format(target))
```
zFernand0: Should we consider moving these up to the core SDK's `validators.py`?

zFernand0 (approved):
LGTM! 😋

zFernand0 (changes requested):
Waiting for functions to be moved to core 🙏

ATorrise:
@zFernand0 moved 🫡

zFernand0 (approved):
LGTM! 😋 
---
Thanks for moving the utility functions to the core sdk 🙏

t1m0thyj (approved):
LGTM, thanks @ATorrise!

PR #404, opened 2026-07-28
Redacted headers and added permissions check

pujal0909:
**What It Does**

Redacts sensitive values from request/response data from logs/exception messages.
Restricts log directory and log file to owner-only access

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [ ] manually tested my changes

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/logger.py
+++ b/src/core/zowe/core_for_zowe_sdk/logger.py
@@ -39,8 +39,7 @@ class Log:
     os.makedirs(dirname, mode=0o700, exist_ok=True)
     os.chmod(dirname, 0o700)
     __log_filename: str = os.path.join(dirname, "python_sdk_logs.log")
-    # Ensure the log file is created with owner-only permissions before any content is written to it,
-    # since it may contain sensitive request/response details.
+
     __log_fd = os.open(__log_filename, os.O_CREAT | os.O_APPEND, 0o600)
     os.close(__log_fd)
     os.chmod(__log_filename, 0o600)
--- a/src/core/zowe/core_for_zowe_sdk/logger.py
+++ b/src/core/zowe/core_for_zowe_sdk/logger.py
@@ -36,8 +36,15 @@ class Log:
     """
 
     dirname: str = os.path.join(os.path.expanduser("~"), ".zowe/logs")
-    os.makedirs(dirname, exist_ok=True)
-    file_handler: logging.FileHandler = logging.FileHandler(os.path.join(dirname, "python_sdk_logs.log"))
+    os.makedirs(dirname, mode=0o700, exist_ok=True)
+    os.chmod(dirname, 0o700)
+    __log_filename: str = os.path.join(dirname, "python_sdk_logs.log")
+    # Ensure the log file is created with owner-only permissions before any content is written to it,
+    # since it may contain sensitive request/response details.
+    __log_fd = os.open(__log_filename, os.O_CREAT | os.O_APPEND, 0o600)
+    os.close(__log_fd)
+    os.chmod(__log_filename, 0o600)
+    file_handler: logging.FileHandler = logging.FileHandler(__log_filename)
     file_handler.setLevel(logging.INFO)
     file_handler.setFormatter(
         logging.Formatter("[%(asctime)s] [%(levelname)s] [%(name)s] - %(message)s", "%m/%d/%Y %I:%M:%S %p")
--- a/src/core/zowe/core_for_zowe_sdk/logger.py
+++ b/src/core/zowe/core_for_zowe_sdk/logger.py
@@ -12,7 +12,34 @@ Copyright Contributors to the Zowe Project.
 
 import logging
 import os
+import subprocess
+import sys
+import getpass
 from typing import Any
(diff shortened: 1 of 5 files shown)


src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -36,8 +36,14 @@ class Log:
...
     dirname: str = os.path.join(os.path.expanduser("~"), ".zowe/logs")
-    os.makedirs(dirname, exist_ok=True)
-    file_handler: logging.FileHandler = logging.FileHandler(os.path.join(dirname, "python_sdk_logs.log"))
+    os.makedirs(dirname, mode=0o700, exist_ok=True)
+    os.chmod(dirname, 0o700)
+    __log_filename: str = os.path.join(dirname, "python_sdk_logs.log")
+
+    __log_fd = os.open(__log_filename, os.O_CREAT | os.O_APPEND, 0o600)
+    os.close(__log_fd)
+    os.chmod(__log_filename, 0o600)
```
zFernand0: Nit:
Given the differences in behavior of the `os.chmod` function, could we explicitly set `follow_symlinks=False`?

Source:
https://docs.python.org/3/library/os.html#os.chmod

---

Also, it seems that Windows  ignores the `exec` bit when `chmod(0o700)`-ing a directory.

src/core/zowe/core_for_zowe_sdk/request_handler.py
```diff
@@ -19,6 +20,30 @@
 from .exceptions import InvalidRequestMethod, RequestFailed, UnexpectedStatus
 from .logger import Log
 
+SENSITIVE_HEADERS = {"authorization", "cookie", "x-csrf-zosmf-header"}
+REDACTED = "<redacted>"
```
zFernand0: I think we don't need to redact the value of the  `csrf` header for z/OSMF since that can be anything. z/OSMF only case that the header is provided.

---

Also, should we try to match the NodeJS SDK's redacted headers?
Image
https://github.com/zowe/zowe-cli/blob/bf84126f0c8368d184a8cbef983823445adcba9b/packages/imperative/src/censor/src/Censor.ts#L40

---

Nit: And instead of using the value `<redacted>`, should we use the same `"****"` as NodeJS for consistency?
Image
https://github.com/zowe/zowe-cli/blob/bf84126f0c8368d184a8cbef983823445adcba9b/packages/imperative/src/censor/src/Censor.ts#L54

src/core/zowe/core_for_zowe_sdk/request_handler.py
```diff
@@ -19,6 +20,30 @@
...
+
+
+def _redact_headers(headers: Any) -> dict[str, Any]:
+    """Return a copy of the given headers with sensitive values redacted."""
+    if not headers:
+        return dict(headers) if headers else {}
```
zFernand0: We are already checking for `if not headers` so no need for the ternary in the `return` statement
```suggestion
    if not headers:
        return {}
```

zFernand0 (changes requested):
LGTM! 😋 
---
Left some comments for consideration 🙏 
I will read up a bit more on the Windows specific behavior of chmod 🤔

src/core/zowe/core_for_zowe_sdk/logger.py
```diff
@@ -36,8 +36,14 @@ class Log:
     """
 
     dirname: str = os.path.join(os.path.expanduser("~"), ".zowe/logs")
-    os.makedirs(dirname, exist_ok=True)
-    file_handler: logging.FileHandler = logging.FileHandler(os.path.join(dirname, "python_sdk_logs.log"))
+    os.makedirs(dirname, mode=0o700, exist_ok=True)
+    os.chmod(dirname, 0o700)
```
zFernand0: Image

Based on that, I don't think this is properly restricting the directory (or the file down below) to be owner-only. They say only the S_IWRITE and S_IREAD will be respected, which means that none of these are set:

Image

Image

Source: https://github.com/python/cpython/blob/3.14/Lib/stat.py

---

We may have to use some ACL commands for windows. For example:
```python
import subprocess
# Grant full access to the current user via icacls
subprocess.run(["icacls", "example.txt", "/grant", "Users:F"])
```

That being the case, can we turn this into a common utility function?

zFernand0 (dismissed):
Can a Windows user confirm the following? 🙏 
@ATorrise @traeok  🙏

src/core/zowe/core_for_zowe_sdk/logger.py
traeok: Your suspicions are correct Fernando, the current call does not handle restricting the directory on Windows to be owner-only. I think we'll need to leverage `icacls` as we did for the Node SDKs to make sure that we properly restrict the permissions of the folder to that user.

traeok (approved):
Thanks Pujal for the fixes, LGTM!

jace-roell (approved):

t1m0thyj (approved):
LGTM, thanks @pujal0909! Tested on Mac and the log permissions look safe 👍

PR #407, opened 2026-07-30
Pin `secrets_core` to v8.35.2

ATorrise:
**What It Does**
Pins the `secrets_core` dependency of the Secrets SDK to the v8.35.1 release tag and saves the commit SHA to cargo.lock. [Dependabot can review if the tag is old and spin up a PR to update to the newest tag and commit](https://github.com/zowe/zowe-client-python-sdk/pull/410) #410 
**How to Test**

**Review Checklist**
I certify that I have:
- [X] updated the changelog
- [X] manually tested my changes

**Additional Comments**

### DIFF
--- a/.github/workflows/secrets-sdk.yml
+++ b/.github/workflows/secrets-sdk.yml
@@ -42,7 +42,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.target }}
-          args: --release --out dist
+          args: --release --locked --out dist
           sccache: "true"
           manylinux: auto
           docker-options: -e PKG_CONFIG_SYSROOT_DIR -e PKG_CONFIG_PATH
@@ -80,7 +80,7 @@ jobs:
         with:
           maturin-version: v1.9.4  # https://github.com/PyO3/maturin/issues/2767
           target: ${{ matrix.target }}
-          args: --release --out dist
+          args: --release --locked --out dist
           sccache: "true"
           working-directory: src/secrets
       - name: Upload wheels
@@ -104,7 +104,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.target }}
-          args: --release --out dist
+          args: --release --locked --out dist
           sccache: "true"
           working-directory: src/secrets
       - name: Upload wheels
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -12,6 +12,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
-# Pinned to a release tag so builds are reproducible
-# Cargo.lock records the exact commit this resolves to
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", tag = "v8.35.1" }
+# Pinned to a commit so builds are reproducible
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" } # v8.35.1
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -13,4 +13,4 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
 # Pinned to an immutable commit so builds are reproducible
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" } # 8.35.1
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "4d2539a49c9469e9293f5fe562f74ec6e874a085" }
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -13,4 +13,4 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
 # Pinned to an immutable commit so builds are reproducible
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "4d2539a49c9469e9293f5fe562f74ec6e874a085" }
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" } # 8.35.1
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -12,4 +12,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", branch = "master" }
+# Pinned to an immutable commit so builds are reproducible
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "4d2539a49c9469e9293f5fe562f74ec6e874a085" }
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -12,5 +12,6 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
-# Pinned to an immutable commit so builds are reproducible
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" } # 8.35.1
+# Pinned to a release tag so builds are reproducible
+# Cargo.lock records the exact commit this resolves to
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", tag = "v8.35.1" }
--- a/src/secrets/Cargo.toml
+++ b/src/secrets/Cargo.toml
@@ -13,4 +13,4 @@ crate-type = ["cdylib"]
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
 # Pinned to a commit so builds are reproducible
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" } # v8.35.1
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "fc8b31ce0de3c7c8a9a371dd43f3daf6a133f002" } # v8.35.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 ### Bug Fixes
 
 - Updated the `pyo3` dependency of the Secrets SDK for technical currency. [#399](https://github.com/zowe/zowe-client-python-sdk/pull/399)
-- Pinned the `secrets_core` dependency of the Secrets SDK to a fixed commit instead of the `master` branch so wheel builds are reproducible, and updated it to Zowe CLI 8.35.1. [#407](https://github.com/zowe/zowe-client-python-sdk/pull/407)
+- Pinned the `secrets_core` dependency of the Secrets SDK to a fixed commit instead of the `master` branch so wheel builds are reproducible.
 
 ## `1.0.0-dev26`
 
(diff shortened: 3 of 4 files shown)


src/secrets/Cargo.toml
```diff
@@ -12,4 +12,5 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", branch = "master" }
+# Pinned to an immutable commit so builds are reproducible
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "4d2539a49c9469e9293f5fe562f74ec6e874a085" }
```
zFernand0: Is there a reason why we want to lock this down to CLI's `8.1.2`?

Image

Image

---

Should we try the latest version which contains a lot of recent fixes? 🙏 
`b2db1840f140b72a3352df56f0802ea4ce213dc5`
https://github.com/zowe/zowe-cli/releases/tag/v8.35.1
Image

```suggestion
secrets_core = { git = "https://github.com/zowe/zowe-cli.git", rev = "`b2db1840f140b72a3352df56f0802ea4ce213dc5`" } # 8.35.1
```

.github/workflows/secrets-sdk.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         uses: PyO3/maturin-action@v1
         with:
           target: ${{ matrix.target }}
-          args: --release --out dist
+          args: --release --locked --out dist
```
zFernand0, line 45: Is `--locked` the equivalent of `--frozen-lockfile` (for yarn and pnpm) and`npm ci`

zFernand0 (changes requested):
LGTM! 😋 
---
Can we use a newer hash (and add the version number as a comment next to it)?

.github/workflows/secrets-sdk.yml
ATorrise: yes :)

src/secrets/Cargo.toml
ATorrise: yes you're right. i think I'm going to make a PR for a bot to make this change periodically that way we don't forget to update - maybe... investigating plausibility

ATorrise: https://github.com/zowe/zowe-client-python-sdk/pull/410

@t1m0thyj @zFernand0 
what do you guys think about this update being done via dependabot

zFernand0 (commented):
LGTM! 😋 
---
waiting on passing checks (which are fixed in #403)

ATorrise:
@zFernand0 hey can you rereview? because i changed to tag comparison so that dependabot can automate the creation of a version-update PR

src/secrets/Cargo.toml
```diff
@@ -12,4 +12,6 @@ crate-type = ["cdylib"]
 
 [dependencies]
 pyo3 = { version = "0.29.0", features = ["abi3-py310", "generate-import-lib"] }
-secrets_core = { git = "https://github.com/zowe/zowe-cli.git", branch = "master" }
+# Pinned to a release tag so builds are reproducible
+# Cargo.lock records the exact commit this resolves to
+secrets_core = { git = "https://github.com/zowe/zowe-cli.git", tag = "v8.35.1" }
```
zFernand0: I do believe dependabot is able to compare commit hashes. I could be wrong though 😅

zFernand0 (dismissed):

src/secrets/Cargo.toml
CBforZ: It can at least for  github actions versions ( like https://github.com/zowe/zowe-cli-mq-plugin/pull/130/changes ) So I would hope it works the same here, but don't know 100% for Cargo

ATorrise: https://github.com/dependabot/dependabot-core/issues/10913
shows this working in a "tag-only" way. Dependabot moved the tag from v1.3.1 to v1.3.2 and the resolved commit in Cargo.lock from ac5664fe to dea187c8. so we only need to watch for new release tags. Cargo resolves the tag to a commit and records it in Cargo.lock, and CI builds with --locked so that exact commit is what compiles. If the lock and the manifest ever disagree the build fails rather than quietly compiling something else.


I think you guys are asking why aren't we just looking at the commit shas? and why decide to do tags to then find the commit sha? There were 49 commits on master between v8.35.0 and v8.35.1, and none touched packages/secrets. Dependabot follows the git ref rather than the crate, so each one is "bump-worthy"

The bigger issue is that watching commits means keeping branch = "master" in Cargo.toml, which names no specific version, so any commit on master is acceptable. The lockfile quietly settles on whichever it resolved first that's how this sat on an Oct 2024 commit that Fernando caught initially in this PR review:
>
> - branch = "master" does not mean "fetch master every build."
> - Cargo only resolves it once. Whenever the crate was first built, cargo fetched master, took whatever the tip happened to be that day, and wrote that exact SHA into Cargo.lock.
> - that was 10 Oct 2024. The tip was 4d2539a when cargo.lock got committed. From that point on every build read the lockfile and used 4d2539a. Cargo never re-checked master, because it had no reason to, so it was pinned all along, just by accident.
> - The manifest said "master," which reads as "we're current." The lockfile said 4d2539a, and the lockfile is what actually decides. 
 
With a tag we get a PR only when zowe-cli actually ships a release, and the commit still gets pinned in Cargo.lock.

zFernand0: Hey Amber, (@ATorrise)

The issue you posted seems to have been opened by accident as the person realized their mistake.

Also, the commit listed in the issue you posted help demonstrate that dependabot can work with git tags, but it doesn't show that dependabot cannot work with commit hashes. There is however a well-written GitHub issue about dependabot only updating the `.lock` file and not the `.toml` file regardless of the method used. 
(_source_: https://github.com/dependabot/dependabot-core/issues/14196)

Lastly, did you use AI to help with your response? If so, please avoid copy-pasting its output when discussing PR review/feedback. I don't believe it's a good use of anybody's time. If you think something this long is necessary, I'd rather just have the conversation in person, over zoom, or a direct message coming from you as opposed to an essay written by AI.

traeok (changes requested):
Is it possible to change this to point to the same commit SHA matching `v8.35.1` contents, rather than specifying a tag? My concern about the tag (along the lines of Fernando's comments): The package is not published, and it is possible to update the tag to point to a new SHA. A commit SHA is immutable and guarantees that no subsequent changes to `secrets-core` *or* the tag will get pulled in as part of build.

src/secrets/Cargo.toml
ATorrise: i used AI to find the link but it is an example of using tags. yes it is made by mistake but that's not the point of the link. 
this is the point: 
>Cargo resolves the tag to a commit and records it in Cargo.lock, and CI builds with --locked so that exact commit is what compiles. If the lock and the manifest ever disagree the build fails rather than quietly compiling something else.

Im trying to explain that from what I understand you can use the tag to find the relevant commit and preserve identity with the exact commit sha. I'm not trying to waste your time im trying to be concise and come with language that is more exact rather than in my simpler way of understanding things

ATorrise: also i heavily worked on that initial response from aggregated context it took me maybe 30-45 minutes to put that all together (the one from 19 hours ago)

zFernand0:
Regarding the safety of git tags...

If a project was to get hijacked and the tag would change to a malicious commit, image

`dependabot` will blindly do a `cargo update` which will result in the update below:
```
Updating secrets_core v0.1.1 (https://github.com/zFernand0/zowe-cli___test-python-sdk?tag=v8.35.1#b2db1840) -> #fc8b31ce
```
And the next CI run `cargo build --release --locked` will just succeed:
```
Compiling secrets_core v0.1.1 (https://github.com/zFernand0/zowe-cli___test-python-sdk?tag=v8.35.1#fc8b31ce)
```

---

And using git-commit hashes:
```
secrets_core = { git = "https://github.com/zFernand0/zowe-cli___test-python-sdk", rev = "b2db1840f140b72a3352df56f0802ea4ce213dc5" }
```

the `Cargo.lock` remains the same (after a `cargo update`) even after the repo is compromised
```
[[package]]
name = "secrets_core"
version = "0.1.1"
source = "git+https://github.com/zFernand0/zowe-cli___test-python-sdk?rev=b2db1840f140b72a3352df56f0802ea4ce213dc5#b2db1840f140b72a3352df56f0802ea4ce213dc5"
```

src/secrets/Cargo.toml
zFernand0: I know we had a brief conversation offline, but I'd love to continue chatting after (or during) lunch 🙏

pujal0909: I think it makes sense to compare git hashes instead of git tags since hashes can't be overwritten and a new one is created anytime there's a change.

So this will just be a tradeoff that we'll have to manually update the git hash since dependabot cannot track/compare hashes in a `Cargo.toml` ? Or is there another automated way to update it ? I guess this is a bit out of scope for this PR.

ATorrise:
@zFernand0 @traeok @CBforZ okay i think i figured out where i was confusing myself

you guys were talking about what happens when a tag gets moved. i was talking about what happens when a new release ships. those are different things and the tag only helps with one of them which i didn't realize until i went and tested it

i thought the tag would give us protection against the sha changing, like dependabot was watching the sha and would tell us if it moved. it isn't. it compares versions and only fires when a higher one shows up. if someone "re-cut" v8.35.1 (i was not thinking this way) onto a different commit, dependabot sees the same name and reports nothing.

the other thing i think i was confusing was what counts as being pinned. basically cargo.lock holds whatever cargo.toml resolves to. so if the tag moved and anyone ran cargo update (the scenario you guys were talking about) the lock would write down the new sha and everything would still look good. so yes you guys are right and we should switch to the sha.

one thing i did find that im not sure anyone anticipated.. i tested this on a fork of the sdk and used the real latest release tag of the cli. dependabot saw the newer tag and opened a normal bump PR. with the rev/sha dependabot made zero requests to zowe-cli and logged:

```
Checking if secrets_core b2db1840... needs updating
Latest version is b2db1840...
No update needed for secrets_core b2db1840...
```
Thinking that I should make a small pr to check for changes (not w dependabot - i was expecting that dependabot was capable of making a sha comparision before the decision to open a pr) so this one stays one line. 

sorry for dragging this out. i was trying to hold onto the automation and didn't understand that i was defending a different thing than what you guys were asking about.

src/secrets/Cargo.toml
ATorrise: @pujal0909 yeah that's what i was trying to hold onto, the part that could make it easier to automate. was trying to figure out how everything could work together so we dont have to manually update. but i have a different pr 410 that is going to do automation outside of dependabot (maybe...)

traeok (approved):
Thanks for updating to use the SHA, LGTM

zFernand0:
Hey Amber,
Sorry for the confusion. And thanks for considering going back to hashes. 🙏 
I have two things I wanted to ask/mention for completeness-sake

---
`#1`
---
Quick question about the testing that wasn't clear to me

> Checking if secrets_core b2db1840... needs updating
Latest version is b2db1840...
No update needed for secrets_core b2db1840...

Did you already have the newer hash? (8.35.2 = fc8b31c ... fc8b31ce0de3c7c8a9a371dd43f3daf6a133f002)
<details> <summary>click here for image ⬇️ </summary>

image

</details>

---
`#2`
---

And one last thing, regarding your observation below:

> There were 49 commits on master between v8.35.0 and v8.35.1, and none touched packages/secrets.

I can't remember exactly the discussion but someone suggested that we could match the hash that last made changes to the secrets-core package,  I think that will be the following:
- `bf84126f0c8368d184a8cbef983823445adcba9b`
- source: https://github.com/zowe/zowe-cli/releases/tag/native-v1.2.7

<details><summary>click here for image ⬇️ </summary>

image

</details>

traeok:
not opposed to using the exact hash that we have for latest secrets-core in the CLI repo, but the current commit used will still yield the latest Secrets SDK used with 8.35.1 - so I feel this is sufficient as-is, but I'm cool with adjusting the SHA

ATorrise:
> Hey Amber, Sorry for the confusion. And thanks for considering going back to hashes. 🙏 I have two things I wanted to ask/mention for completeness-sake
> 
#1
no still on b2db1840 (v8.35.1). fc8b31ce was already published and available upstream, it just never proposed it. the only difference of the fork was the change between tag vs rev. with the tag it found v8.35.2 and opened the bump PR. with the rev it made zero requests to zowe-cli at all and finished in 1ms. that's why the "Latest version is b2db1840" line is weird. it's calling the commit we pinned the latest version, because with a rev there's no version to compare against so the pinned sha is latest by definition. it's not that it looked and found nothing, it never looked.

#2
i ran `git diff b2db1840 bf84126 -- packages/secrets`. the diff is empty so these commits build identical code, but since v8.35.2 shipped while this was open, ill bump to that release's sha (fc8b31)instead of bf84126.

zFernand0 (approved):

PR #408, opened 2026-07-30
gene's encode_uri added/adapted to python

ATorrise:
**What It Does**

Ports the URI-encoding fixes from Gene's Zowe CLI PR zowe/zowe-cli#2758 to the Python SDK, adding EncodeUri logic
- Mirrors the 2-pathed encoding differences/strictness between uss (`_encode_uri_path_for_uss`) and ds(with ds being less restrictive)(`_encode_uri_path_for_zos`)
- Includes bug fix found when porting: 
  - `Console.__init__` put the literal string "defcn" (the default console name) right into the base URL, then str.replace replaced every instance of defcn in the string, so if for some reason there was a valid defcn in the URL, that would get replaced and break the URL
 
**How to Test**

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/sdk_api.py
+++ b/src/core/zowe/core_for_zowe_sdk/sdk_api.py
@@ -11,6 +11,7 @@ Copyright Contributors to the Zowe Project.
 """
 
 import copy
+import posixpath
 import urllib
 
 from . import session_constants
@@ -18,6 +19,24 @@ from .logger import Log
 from .request_handler import RequestHandler
 from .session import ISession, Session
 from typing import Any, Optional, Type
+
+# Characters that fail against both z/OSMF and API-ML unless they are encoded.
+_USS_CHARS_TO_ENCODE = {" ": "%20", "%": "%25", "+": "%2B", "?": "%3F"}
+
+# Characters that API-ML rejects with an HTTP 400 unless they are encoded.
+# None of these are encoded for a direct z/OSMF connection.
+_APIML_CHARS_TO_ENCODE = {
+    "#": "%23",
+    ";": "%3B",
+    "<": "%3C",
+    ">": "%3E",
+    "[": "%5B",
+    "]": "%5D",
+    "^": "%5E",
+    "{": "%7B",
+    "|": "%7C",
+    "}": "%7D",
+}
 
 
 class SdkApi:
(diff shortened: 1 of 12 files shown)


src/core/zowe/core_for_zowe_sdk/sdk_api.py
```diff
@@ -112,3 +131,91 @@ def _encode_uri_component(self, str_to_adjust: str) -> str:
...
+            if next_char == '"':
+                # Both encoded and unencoded double-quotes fail in REST requests
+                self.logger.error(f"The USS path '{uss_uri_path}' contains a double-quote character.")
+                raise ValueError(
+                    f'The supplied USS path \'{uss_uri_path}\' contains a double-quote " character. '
+                    "When a double-quote is present, z/OSMF and API-ML servers fail with an HTTP 400 "
+                    "or 500 error code. This request was not sent."
+                )
+            if next_char in _USS_CHARS_TO_ENCODE:
+                encoded_path.append(_USS_CHARS_TO_ENCODE[next_char])
```
t1m0thyj, line 216: Since `%` is included in the map of `_USS_CHARS_TO_ENCODE`, if a URI contains characters that are already encoded such as `%20` for space, it will get converted to `%2520` which is not valid.

Not sure if it's in scope to fix in this PR, given that the implementation in the Node SDK seems to have the same problem, and the input passed to this method should typically not be encoded yet. Thoughts @zFernand0 @traeok?

traeok: Would this mean that merging would cause a regression in behavior given the scope of expected URIs?
I think we could address this separately since the Node.js SDK already suffers from the same problem - but since we've now identified it can be an issue, maybe we can proactively add a note for developers to clarify that pre-encoded URIs are not accepted (until this is resolved).

awharn (approved):

traeok (approved):

src/core/zowe/core_for_zowe_sdk/sdk_api.py
traeok: Filed this as a separate issue (we can label as low priority) https://github.com/zowe/zowe-client-python-sdk/issues/413

PR #411, opened 2026-08-07
Fix secure credential loading for nested team-config profiles

ATorrise:
**What It Does**
When testing https://github.com/zowe/zowe-client-python-sdk/pull/408 i found that the way i structured my config was returning 401s
ie: 
```
mainframe ->
  -> properties
       -> host
  -> profiles
       -> zosmf ->
            -> properties
                 -> port
            -> secure
                 -> user
                 -> password
```   
                 
found that `__load_secure_properties` looked for zosmf as a direct key of the mainframe profile but zosmf actually lives one level deeper, under the profiles key.

the fix to config_file.py handles however many levels a config nests 

`test_custom_file_and_custom_profile_loading_with_nested_profil`e user/password values came from a plaintext override in `nested.zowe.config.user.json`, not from the vault. So apparently no test ever exercised `__load_secure_properties` for a profile nested more than one level deep with secure fields on the child! Added a new test (`test_nested_profile_with_secure_properties_on_child`) that pulls creds from the mocked vault
**How to Test**
change your config to match this level of nesting when testing. verify that you get 401s on main and you can run system tests on this branch :)
**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/config_file.py
+++ b/src/core/zowe/core_for_zowe_sdk/config_file.py
@@ -383,7 +383,7 @@ class ConfigFile:
         for k, v in profiles.items():
             if not isinstance(v, dict):  # Ensure v is a dictionary
                 if not self.__suppress_config_file_warnings:
-                    self.__logger.warning("Invalid profile passed when schame validation is off")
+                    self.__logger.warning("Invalid profile passed when schema validation is off")
                 continue  # Skip invalid entries
 
             if segments[0] == k:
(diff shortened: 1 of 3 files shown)


zFernand0 (dismissed):
LGTM! 😋 
---
Very good catch! 🙏 

Here is a quick before and after of this fix:

**Before**
```
{'rejectUnauthorized': False, 'host': 'lpar.1', 'port': 1234}
HTTP Request has failed with status code 401. 
```

**After**
```
{'rejectUnauthorized': False, 'host': 'lpar.1', 'port': 1234, 'user': 'dummy', 'password': 'dummy'}
HTTP Request has failed with status code 401. 
```
I got the 401 because `dummy:dummy` is not valid on my system

---
With proper creds:
```
{'rejectUnauthorized': False, 'host': 'lpar.1', 'port': 1234, 'user': 'REALUSER', 'password': 'PASSWORD'}
//* //FERNAN00 JOB (IZUACCT),FERNANDO,CLASS=B,MSGCLASS=X,TYPRUN=HOLD
...
```

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -455,6 +455,8 @@ def __load_secure_properties(self) -> None:
                     if i == len(segments) - 1:
                         profiles_obj.setdefault("properties", {})
                         profiles_obj["properties"][property_name] = value
+                    else:
+                        profiles_obj = profiles_obj.get("profiles", {})
```
traeok: I know the loop is pre-existing, but this walk looks identical to `find_profile`. Could we just use that instead and remove this whole loop?

For example:
```py
profile = self.find_profile(".".join(segments), self.profiles)
if isinstance(profile, dict):
    profile.setdefault("properties", {})[property_name] = value
```

traeok (commented):
I noticed that even with these changes, the failure is never raised from `check_missing_props`, which is likely why @zFernand0 was seeing a 401 rather than an explicit error. I'd consider it out of scope for this PR, but worth a follow-up issue.

Aside from that, I left a suggestion around the modified code in an attempt to de-duplicate logic.

src/core/zowe/core_for_zowe_sdk/config_file.py
zFernand0: resolved in c9f60d1

image

zFernand0: I often like to run some sanity checks with any changes I make.

<details><summary>Here is a brief sumary: (click here) ⬇️ </summary>
image
</details>

zFernand0:
The missing coverage is just on a line that I fixed a typo 😢

awharn (approved):

traeok (approved):
I thoroughly reviewed c9f60d1 in addition to the PR as a whole, no concerns here and code changes LGTM 😋 
The only outstanding concern is that patch is at 75% instead of the target 80%, but since its not a required stage for Python SDK I'm not going to hold up the PR over it.

Thanks Fernando!

PR #412, opened 2026-08-07
Remove support for schemas that are loaded from a URL (breaking)

ATorrise:
**What It Does**
Removes support for loading a JSON schema from a remote `http(s)://` URL via a config file's `$schema` property.

- `validate_config_json` (used during profile loading) now raises a `ValueError` if `$schema` is an `http(s)://` URL, telling the user to use a local file path instead.
- `ConfigFile.schema_list` (used by `get_env`) now returns an empty list and logs/warns instead of making a network request for a remote `$schema` URL.

Only local schema files (absolute path, relative path, or `file://`) are supported now.

**How to Test**
run the new tests and make sure they pass 

**Review Checklist**
I certify that I have:
- [x] updated the changelog

**Additional Comments**
Breaking! Cases where `$schema` points at a remote url now fail instead of silently succeeding

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 ### Bug Fixes
 
-- Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported.
+- **Breaking**: Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported.
 - Redacted request headers and restricted log directory/file to owner-only access. [#404](https://github.com/zowe/zowe-client-python-sdk/pull/404)
 - Fixed `Jobs.get_job_output_as_files` writing to a directory it never created, and made job output paths stay within the target directory. [#403](https://github.com/zowe/zowe-client-python-sdk/pull/403)
 - Updated the `pyo3` dependency of the Secrets SDK for technical currency. [#399](https://github.com/zowe/zowe-client-python-sdk/pull/399)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 ### Bug Fixes
 
-- **Breaking**: Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported.
+- **Breaking**: Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported. [#412](https://github.com/zowe/zowe-client-python-sdk/pull/412)
 - Redacted request headers and restricted log directory/file to owner-only access. [#404](https://github.com/zowe/zowe-client-python-sdk/pull/404)
 - Fixed `Jobs.get_job_output_as_files` writing to a directory it never created, and made job output paths stay within the target directory. [#403](https://github.com/zowe/zowe-client-python-sdk/pull/403)
 - Updated the `pyo3` dependency of the Secrets SDK for technical currency. [#399](https://github.com/zowe/zowe-client-python-sdk/pull/399)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ All notable changes to the Zowe Client Python SDK will be documented in this fil
 
 ### Bug Fixes
 
+- **Breaking**: Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported.
```
zFernand0:
```suggestion
- **Breaking**: Removed support for loading a JSON schema from a remote `http(s)://` URL via the `$schema` config property. Local schema files are still supported. [#412](https://github.com/zowe/zowe-client-python-sdk/pull/412)
```

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -180,15 +179,11 @@ def schema_list(self, cwd: Optional[str] = None) -> list[dict[str, Any]]:
...
-                response.raise_for_status()  # Ensure it's a valid response
-                schema_json = response.json()
-            except requests.RequestException as e:
-                if not self.__suppress_config_file_warnings:
-                    warnings.warn(f"Invalid schema request: {e}")
-                    self.__logger.warning(f"Invalid schema request: {e}")
-                return []
+            # remote schema loading is not supported
+            if not self.__suppress_config_file_warnings:
+                warnings.warn(f"Loading a JSON schema from a remote URL is not supported: {schema}")
```
zFernand0: We seem to be using this same string in multiple paces... wondering if we could move it somewhere else and reuse it.

src/core/zowe/core_for_zowe_sdk/config_file.py
```diff
@@ -180,15 +179,11 @@ def schema_list(self, cwd: Optional[str] = None) -> list[dict[str, Any]]:
...
+                self.__logger.warning(f"Loading a JSON schema from a remote URL is not supported: {schema}")
+            return []
```
zFernand0: I think we should be raising the error regardless.

Right now, the only place that's calling the `schema_list` function is inside the `get_env` function, which is only called when the user specifies that they want to `override_with_env`.

If we return an empty list here, the developer/user may not realize that their environment variables were not loaded because the schema comes from a URL. This leads to unexpected behavior.

zFernand0 (dismissed):
LGTM! :yum:
---
I do have one small request about raising the error.
Feel free to get other developer's opinion on this.

src/core/zowe/core_for_zowe_sdk/config_file.py
zFernand0: resolved in 952871af8b12f50b886ce10fe13b46626d575376
Screenshot 2026-08-24 at 16-16-07

src/core/zowe/core_for_zowe_sdk/config_file.py
zFernand0: resolved in 2822a86

zFernand0 (commented):
LGTM! 😋

CBforZ:
i think this test is still expecting a warning instead of an error: 

```
            with self.assertWarns(UserWarning):
>               result = config_file.schema_list()

tests/unit/core/test_profile_manager.py:307: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
```

tests/unit/core/test_profile_manager.py
```diff
@@ -296,6 +296,19 @@ def test_validate_schema_logger(self, get_pass_func, mock_logger_warning: mock.M
...
+        with mock.patch("requests.get") as mock_get:
+            config_file = ConfigFile(
+                name="zowe_abcd", type="User Config", schema_property="https://example.com/zowe.schema.json"
+            )
+            config_file.suppress_config_warnings(False)
+            with self.assertWarns(UserWarning):
+                result = config_file.schema_list()
+
+        mock_get.assert_not_called()
+        self.assertEqual(result, [])
```
traeok: This test expects a `UserWarning` and an empty `result` list, but `config_file.py` (lines 186-188) raises a `ValueError`

traeok (dismissed):
Seeing a desync between the test and latest code changes, plan to approve once resolved 😋

tests/unit/core/test_profile_manager.py
zFernand0: resolved in fbfccf7

CBforZ (approved):

traeok (approved):
LGTM Thanks @zFernand0 for the quick resolution 😋

PR #415, opened 2026-09-04
`_encode_uri_path_for_uss` acknowledge encoding on a URI given as input

### LINKED ISSUE #413: `_encode_uri_path_for_uss` does not skip encoded URIs
**Describe the bug**

`_encode_uri_path_for_uss` function does not acknowledge encoding on a URI given as input, which can result in invalid URI encoding sequences.

**Expected and actual results**

**Expected:** `_encode_uri_path_for_uss` should either skip encoded URIs, or at least have a note on the function clarifying that it should not be used with a URI that's already encoded.
**Actual:** The function encodes the URI again, causing incorrect URI encoding.

Example Python logic to detect and return early for this case:

```py
from urllib.parse import unquote

def is_uri_encoded(uri: str) -> bool:
    """Returns True if the URI contains percent-encoded characters."""
    return uri != unquote(uri)

# Examples
print(is_uri_encoded("https://example.com%20world"))  # True
print(is_uri_encoded("https://example.com world"))  # False
```

aadityasinha-dotcom:
**What It Does**
Resolves #413 

**How to Test**

**Review Checklist**
I certify that I have:
- [ ] updated the changelog
- [ ] manually tested my changes

**Additional Comments**

### DIFF
--- a/src/core/zowe/core_for_zowe_sdk/sdk_api.py
+++ b/src/core/zowe/core_for_zowe_sdk/sdk_api.py
@@ -12,6 +12,7 @@ Copyright Contributors to the Zowe Project.
 
 import copy
 import posixpath
+import re
 import urllib
 
 from . import session_constants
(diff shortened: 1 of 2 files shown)


traeok (dismissed):
While I'm not opposed to taking this approach, I noticed that `urllib` is already imported at the top. Could we use the `urllib.parse` function to detect URI encoding instead?

For example:

```py
from urllib.parse import unquote, quote

def is_percent_encoded(text: str) -> bool:
    """Checks if a string contains valid percent-encoding."""
    # unquote transforms '%20' to ' ' and returns a different string if encoding was present
    return unquote(text) != text
```

traeok (commented):
Thanks Aaditya, all looks good and I'm ready to approve if tests pass. Can you please add a changelog entry to track the fix?

